PR #1, opened 2015-06-25
Feature/import satellite

hellais:
With @willscott we are adding support for importing satellite DNS scans to the ooni-pipeline.


gabelula:
This is a very very old pull request. Can we close it?

darkk:
@gabelula, yep, I don't think that Sattelite import is something that is going to happen soon. I also don't see any Satellite-specific code here, so I see no reason to keep the PR open.

PR #2, opened 2015-11-25
Feature/docs

hellais:
Documentation for the ooni-pipeline tasks


hellais:
@joelanders can you see if this makes sense and if so sign off on it?

PR #3, opened 2015-11-29
Fixes to support new versions of luigi and aws. Explicitly install gcc and python-devel

hellais:
Replace ooni-pipeline-ng with ooni-pipeline


joelanders:
sieht schön aus

PR #7, opened 2015-12-11
Joe/redo

joelanders:
simplified, clarified upload_reports and sync_reports.
mostly by removing bits that called other tasks (in a non-luigi-dependency kind-of-way), renaming things.

those two tasks are now called `move_and_bin_reports` and `move_reports`.

`move_reports` is meant to move reports from a collector to the incoming bucket.
`move_and_bin_reports` moves reports from incoming to private/raw_reports and renames/folders them. (tomorrow i can make it folder them based on upload date rather than probe date).

i'm still thinking how to fit our problem to the luigi model: specifically, how do the "aggregate" and "insert into table" tasks know what's been processed already.

we could bin into folders by upload date, and then _promise_ never to add things to a folder after we've run the "aggregate files within a folder" task... seems like there should be something cleverer / less fragile. i want to be able to ask "are all the raw reports accounted for in the final products?" without redoing all of the processing work. postgres would let you query it for report_ids. maybe if the json streams had a header listing the report_ids inside.

zzz


hellais:
Very good work. I imagine this will therefore super-seed #6?

hellais:
I see that the `halt` option has been removed, that needs to be present in order to stop the ephemeral ec2 machine once it runs the operations on it.

joelanders:
yes and yes. i imagined the halting ought to be done by whomever calls the invoke script.
`invoke blah && halt`

hellais:
The halt needs to be run inside of the ephemeral VPS though, since that is the instance we want to shut down (the one started with the start_computer task), hence it needs to be inside the task itself.

joelanders:
hm, i see how it's kinda needed by the current workflow, but it feels like a hack.

hellais:
I think we may have to re-run it since there were some problems with the poping of the response body.

Also the `PRIMARY KEY` is set to be the tuple (input, report_id) it will fail to insert into the database values that have the `input` set to NULL (there are many like this) since a PRIMARY KEY is a CONSTRAINT `UNIQUE` and `NOT NULL`.
I think that because of this we should create another ID that we use as the PRIMARY KEY and set it to be generated automatically like so:

```
id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
```

PR #12, opened 2016-02-21
Add mappings for censorship circumvention tool tests

hellais:


hellais:
I am going to proceed in merging this.

PR #14, opened 2016-03-03
Move update views into separate tasks

hellais:


hellais:
I am going to proceed in merging this since it's been working deployed on the pipeline since days now.

PR #16, opened 2016-03-12
Add support for detecting blockpages in India

hellais:


pipeline/batch/sql_tasks.py
```diff
@@ -147,6 +148,24 @@ def identified_vendors(metrics_table):
                 LIKE '%squid%'
             OR {metrics_table}.test_keys -> 'received' ->> 3
                 LIKE '%squid%')
+UNION
```
anadahz, line 151: Is this addition really related to the identification of IN block pages?

hellais, line 151: Actually no, this is related to privoxy detection. I guess the commit message is indeed a bit misleading.

PR #20, opened 2016-04-26
Feature/vagrant

hellais:


anadahz:
Should this be 2 separates PRs or are both related to Vagrant?

hellais:
I think this can be closed.

Long live the new pipeline!

PR #22, opened 2016-05-14
Fingerprint/qa kr

hellais:
This has blockpages for KR and QA


anadahz:
This looks good merging!

PR #37, opened 2016-08-10
Add new blockpage for korea

hellais:


anadahz:
@hellais the KR blockage fingerprint has been already added here: https://github.com/TheTorProject/ooni-pipeline/blob/1333c5c2b2c503aa732342a71f4c970a9f1f90c8/pipeline/batch/sql_tasks.py#L32

hellais:
@anadahz that is a different one, it's in the Location header, this is in the body.

anadahz:
@hellais are these block pages on a different ASN that does not include the Location header?
I'm afraid that this fingeprint could trigger some false positives.

hellais:
> @hellais are these block pages on a different ASN that does not include the Location header?

Yes.

> I'm afraid that this fingeprint could trigger some false positives.

I have not seen any false positives in the results collected so far, if you see some let me know.

anadahz:
Not really false positive but some interesting cases, perhaps web connectivity false negative tests:
https://explorer.ooni.torproject.org/measurement/20160618T063503Z_AS3786_TOpjS3B3C4hNYmSKOOrhTPJkqKjZ6UhbevcD7uGkSZfH5Bm6gR?input=http:%2F%2Fwww.mrchewsasianbeaver.com
https://explorer.ooni.torproject.org/measurement/20160810T072625Z_AS4766_vJqT3pT9IkPUF35sgNGr3LWUaxlCl9Z7W8MgcHk7QhWijv63VK?input=http:%2F%2Fwww.kcna.kp
https://explorer.ooni.torproject.org/measurement/20160418T210843Z_fqVeZsJAOPiTAwELwVeOAntlyGlIfsnrhONaNmIhEsMHTOcQzF?input=null
https://explorer.ooni.torproject.org/measurement/20160418T211041Z_fqVeZsJAOPiTAwELwVeOAntlyGlIfsnrhONaNmIhEsMHTOcQzF?input=null
https://explorer.ooni.torproject.org/measurement/20160418T211224Z_fqVeZsJAOPiTAwELwVeOAntlyGlIfsnrhONaNmIhEsMHTOcQzF?input=null
https://explorer.ooni.torproject.org/measurement/20160812T071942Z_AS4766_KAgt4GeVuVkZngDsZNyMYDZ6soVO2FEjy1s756s3IoluNd08ZK?input=null

hellais:
Yes, though those are being flagged properly by the blockpage detection (you will see that they do appear in the country page).

The reason why there is this mis-alignment in the client side and server side heuristic is explained here: https://github.com/TheTorProject/ooni-explorer/issues/80#issuecomment-204391373

PR #38, opened 2016-08-11
Update ooni-pipeline-architecture

hellais:
In this branch I am going to update the architecture documentation for the current and future ooni pipeline.


ghost:
Where would a good place be for architecture proposals? Would that be something that should be consolidated to IRC, or made open for discussion on ooni-dev?

hellais:
@TylerJFisher it depends at what stage the proposal is at. If it's in the form of "I have these ideas and would like to know what you think" then you could probably post it to ooni-dev or you could even suggest in a document as a pull request to this repo.

I am still finishing spec-ing out the current design and was going to then drop some notes on what I would like to see improved in the next version.

hellais:
I am going to proceed in merging this.

PR #42, opened 2016-09-15
Create a multicolumn index on test_start_time and probe_cc

hellais:
This leads to a significant performance boost to queries on countries
with few measurements (query time for common queries went from 396000ms to 300ms)


hellais:
Merging since this is working well.

PR #43, opened 2016-10-02
Refresh materialized views concurrently

hellais:
Prior to this materialized view updates lead to the database tables
related to these views will be locked (also for read), which makes the
explorer UI irresponsive during the 1 hour or so that the materialized
views are being updated.
This makes it happen concurrently, which means that the updated view is
written to a temporary table and then renamed in place, making the old
table still accessible while the update is in progress.

### DIFF
--- a/pipeline/batch/daily_workflow.py
+++ b/pipeline/batch/daily_workflow.py
@@ -844,7 +844,7 @@ class UpdateView(RunQuery):
 
     view = ''
     def query(self):
-        return 'REFRESH MATERIALIZED VIEW {view};\n'.format(view=self.view)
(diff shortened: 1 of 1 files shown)


darkk:
Do we need to turn AUTOVACUUM on in this case?

hellais:
@darkk good point. I remember reading that when I looked at the docs in the past, but forgot it while doing this. Will check with the docs and see if that is true and if so also enable that.

hellais:
Ok so I looked into this and it seems like the CONCURRENT keyword requires you to VACUUM after: https://wiki.postgresql.org/wiki/What%27s_new_in_PostgreSQL_9.4#REFRESH_MATERIALIZED_VIEW_CONCURRENTLY

hellais:
I am going to merge this, but we should be sure to enable AUTOVACUUM on the database.

PR #44, opened 2016-10-13
Add psycopg2 to requirements.txt #1

andresazp:
I believe psycopg2 is a requirement, probably should be on requirements.txt


anadahz:
Thanks, psycopg2 is indeed a requirement of current ooni-pipeline.

Issue #61, opened 2017-05-03
Data Analysis: Core Questions to Answer


agrabeli:
Every time I write OONI research reports, I always have a set of basic questions that I ask. These are not sophisticated/complicated research questions, but rather basic questions that can help derive basic meaning from the network measurement data. 

I think these questions should be considered as part of the engineering of OONI's data processing pipeline. These questions include:

1. **Which is the testing period?** When did the network measurement collection start for country X and when did it end?

2. **Which is the data analysis period?** The network measurements are analysed from which date until which date? (it might not take into account all of the network measurement data that has ever been collected from country X)

3. **From which ASNs have network measurements been collected in country X?** A list of all ASNs per country, as well as the data collected from each ASN, would be useful.

4. **Which OONI tests were run in country X?** And in which ASN was each OONI test run in, and during which time period?

5. **The measurements of which OONI tests are considered as part of this data analysis?** (e.g. Web Connectivity, HTTP Invalid Request Line, HTTP Header Field Manipulation, Vanilla Tor, etc.)

6. **As part of data analysis, did you detect the presence of any middle boxes?** If so, when and in which networks?

7. **As part of data analysis, did you find Tor, WhatsApp, Facebook Messenger, or Telegram to be blocked?** If so, when and in which networks?

8. **Analyzing Web Connectivity measurements for country X:** 

- Which sites served block pages (and in which ASNs)? (if any)

- Which sites presented anomalies? A list, categorizing the anomalous sites from the most anomalous to the least anomalous, would be useful.

- How many anomalies did each site present? 

- What types of anomalies did each site present? (e.g. TCP/IP, DNS, HTTP-diff, HTTP-failure)

- How frequently did these sites present network anomalies across the testing period? A timeline of anomalies per site would be useful.

- How many times was an anomalous site tested overall? A site presented X anomalies out of Y times it was tested.

- When did these sites start (and stop) presenting network anomalies?

- Where (in which ASN) did these sites present network anomalies? A comparison of anomalous sites across ASNs would be useful.

- Can the accessibility of anomalous sites be compared with access to them from a control vantage point? A site, for example, may present a high amount of anomalies, but that may not be due to censorship. Instead, that might be because the site is down or because the site's server is generally quite flakey.

By integrating the above as part of the engineering of OONI's data processing pipeline, I think we will be able to derive more meaning from the network measurement data.

agrabeli:
[...]

Issue #32, opened 2016-07-06
Redesign the data processing pipeline


hellais:
We should write a design document for how the next iteration of the data processing pipeline.

This design document should feature:
- How data is ingested into the pipeline (how is data published to it, how does it fetch data from third parties)
- What are the various data processing stages that it should go through
- How it is analysed and post processed for being consumed by the explorer
- What are the output data products of it
- What technological choices should be made for the various software components of it

On the last point I have looked a bit into airflow (https://github.com/apache/incubator-airflow). I quite like it's model and would be interested in considering it as a workflow engine to replace luigi.

What I like very much about it are:
- It has it's own scheduler and doesn't require us to depend on cron for triggering the period tasks
- It supports using celery based message queues for scheduling tasks to be run (http://pythonhosted.org/airflow/concepts.html#queues)
- The operators support templating via jinja2 and there is a `BashOperator` that is quite flexible

cc @willscott

willscott:
Airflow looks like it can help organize a bunch of the complexity in the current structure! The design points to make sure we can do cleanly should include:
- How does a submission get into the pipeline? If we have a web hook for submission, can that go in a message queue? what does that ingestion look like?
- If we want regular incremental updates, how well does airflow do? we'll eventually want at least hourly updates of some products.

ghost:
@hellais @willscott: to expand on this a little bit, I can definitely see how Airflow's trigger / hook / sensor routines could be used to improve the overall flexibility of `ooni-pipeline`. 

Pros: 
- Apache Airflow provides its own schedule which opens up the possibility of autonomously triggering time-based jobs (e.g. back-up, archival, and reporting);
- Apache Airflow provides sensors, hooks, and triggers which could be used to build an event-driven pipeline;
- Apache Airflow would open up the possibility of submitting `ooni-probe` test results directly from collectors given the presence of sensors / hooks / triggers;
- The technical complexity associated with developing new operators seems to be minimal; and
- The user community is pretty active:
  - https://groups.google.com/forum/#!forum/airbnb_airflow
  - http://mail-archives.apache.org/mod_mbox/incubator-airflow-dev/201606.mbox/browse

I'd like to propose the following points on top of what's already been listed:
- How is the intermediate state of test data within each pipeline phase going to be maintained?
- What intermediate format should test data be in between each pipeline phase?
- Should the pipeline be distributed, or should it be consolidated to one node?
- What were the bottlenecks associated with the previous iteration?
[...]

PR #62, opened 2017-05-08
Stable ooni-pipeline 3.0.0

darkk:
Here is PR to track overall progress of epic pipeline refactoring #32, it's **not ready** for merging. 
The checklist below is living "definition of done" of the epic as captured in `DoD-16.10.txt` :-)

- [x] Serve reports from HKG to AMS
  - [x] Split large reports -- no good reason to do that yet
  - [x] measure latency & bandwidth between AMS and HKG
  - [x] understand "batch" size to be able to fetch it within 1 RTT -- cwnd=70…80, RTT=260ms
  - [x] check if snappy/lz4/lzo provide proper API to create a "seekable" stream -- they do
  - [x] merge tiny reports
  - [x] verify that 90kb can be really fetched using keep-alive within 1 RTT
    - for i in `seq 35`; do curl -sS -w 'time_total:  %{time_total}\n' -o /dev/null -H 'Range: bytes=0-1550000'  http://127.0.0.1/64M; sleep 2 ; done
    - bytes=0-1600000 is bad & net.ipv4.tcp_slow_start_after_idle=0
    - plain `ip route add … initcwnd 1200` does not reduce 1st req latency (but that's fixable), sort of "Persist cwnd" at https://www.chromium.org/spdy/spdy-best-practices
  - [x] find optimal batch size for SANITIZED reports
  - [x] not tested https://github.com/facebook/zstd also unclear patents (as well as with React!): https://github.com/facebook/zstd/issues/335 & https://news.ycombinator.com/item?id=12692552
  - [x] test throughput and latencies between Greenhost VMs while reading data from NFS datastore and "fake S3-like read-only store", (NFS share exported via HTTP using nginx) to have a baseline & estimate penalty for transmission and decompression; use these numbers to estimate chunk size

- [x] python-lz4
  - [x] some API calls do not return errors on unknown/bad input, so it's not "typo-safe"
  - [x] no value in pydoc lz4.frame ```auto_flush (int): Enable () or disable (0) autoFlush```
  - [x] compression API is not "pythonic", e.g. context is not an object with methods, it's just a pointer that has to be passed to other functions
  - [x] decompression API is incomplete, e.g. it does not support frame concatenation by any means: either returning an error if there is some non-consumed data or returning amount of consumed data
  - [x] replace lz4 Popen with python-lz4 + on-the-flight crc32 checking
  - [x] context leaks memory if free_compression_context() is not explicitly called
  - [x] there are minor documentation typos
  - [x] why is it good? there was hypothesis, that response body takes significant amount of bytes, but base64() is usually badly compressible with lz4
  - [ ] streaming decompress()
  - [x] mem leak: valgrind --leak-check=full python -c 'import lz4.frame as f; a = f.create_compression_context(); print a; del a' see also python-lz4/python-lz4#10
  - [x] no EPARAM: python -c 'import lz4.frame as f; print f.create_compression_context("whalam", ba="lamba")' see also python-lz4/python-lz4#11
  - [x] python-lz4/python-lz4#12

- [ ] data mismatches
  - [x] TZ mismatch in datetime & bucket date
        ooni/public/sanitised/2016-08-01/20160731T104202Z-US-AS14618-dns_injection-no_report_id-0.2.0-probe.json
        ooni/private/reports-raw/2016-08-01/20160731T104202Z-US-AS14618-dns_injection-no_report_id-0.2.0-probe.json
  - [ ] no sanitiesed data: ooni/private/reports-raw/2016-05-17/20160516T070552Z-GB-AS5607-http_requests-no_report_id-0.1.0-probe.yaml
  - [ ] outliers -- **measurements** with 13mb of COMPRESSED data, 2014-11-22 with ~1M measurements
  - [ ] `2017-02-14` bucket has no "benign mismatches" log stored, consider recomputing
  - [ ] partial bucket processing and re-processing to be tolerant to issues like #65 

- [ ] infrastructure
  - [x] airflow 1.7.2 stabilization
  - [x] secure postgresql/rabbitmq/FERNET_KEY with non-default values
  - [x] airflow 1.8.0 update & deployment playbook
  - [x] secure airflow docker VMs from outer world
  - [x] deployment based on static images from http://hub.docker.com/ instead of building the image on every host
  - [ ] set up AMS replica of HGK metadb
  - [x] `ntp` and poweroff safety, TheTorProject/ooni-sysadmin#101
  - [ ] consider backups / replicas of metadata DBs

- [ ] data "massage"
  - [x] cleanup `reports-raw` that are `canned` & backed in s3 to save disk space
  - [x] cleanup `sanitized` that have matching `autoclaved` files
  - [ ] scripts to manage redundant copies of data in Amazon S3
  - [ ] check mismatches between `sanitized` and `autoclaved`
  - [ ] re-run whole pipeline as `shovel:0.0.6` had a bug possibly producing duplicates at features table on two successful executions of centrifugation.py with same arguments (no idempotence here)
  - [ ] check if any reports/measurements were lost due to `JOIN` instead of `LEFT JOIN` on metadata tables

- [ ] More ETL & database aggregation logic
  - [ ] #61 
  - [x] faster 64bit simhash computation using [seomoz/simhash-cpp](https://github.com/seomoz/simhash-cpp)
  - [ ] simhash-based blockpage detection
  - [ ] DNS `A` records-based blockservers detection
  - [ ] stats for .annotations.{platform,test_class} (requested by @hellais) 

- [ ] shutdown pipeline based on `luigi` and replace it with this one, based on `airflow`
  - [x] temporary move old DB to separate SSD-based box


darkk:
`openobservatory/shovel:0.0.6` brings some new bugs:

- [x] postgres detects deadlock on three concurrent writes to `input` and `residual` tables
<details> 
  <summary>ERROR:  deadlock detected ...</summary>

```
ERROR:  deadlock detected
DETAIL:  Process 1412 waits for ShareLock on transaction 1540; blocked by process 1405.
 Process 1405 waits for ShareLock on transaction 1551; blocked by process 1412.
 Process 1412: 
             INSERT INTO input (input)
             SELECT DISTINCT input FROM measurement_blob_ WHERE input IS NOT NULL
             ON CONFLICT DO NOTHING
         
 Process 1405: 
             INSERT INTO input (input)
             SELECT DISTINCT input FROM measurement_blob_ WHERE input IS NOT NULL
             ON CONFLICT DO NOTHING
         
HINT:  See server log for query details.
CONTEXT:  while inserting index tuple (1411,93) in relation "input"
STATEMENT:  
             INSERT INTO input (input)
             SELECT DISTINCT input FROM measurement_blob_ WHERE input IS NOT NULL
             ON CONFLICT DO NOTHING
```
and
```
ERROR:  deadlock detected
DETAIL:  Process 2190 waits for ShareLock on transaction 2332; blocked by process 2208.
 Process 2208 waits for ShareLock on transaction 2316; blocked by process 2190.
 Process 2190: 
             INSERT INTO residual (residual)
             SELECT DISTINCT residual FROM measurement_blob_
             ON CONFLICT DO NOTHING
         
 Process 2208: 
             INSERT INTO residual (residual)
             SELECT DISTINCT residual FROM measurement_blob_
             ON CONFLICT DO NOTHING
         
HINT:  See server log for query details.
CONTEXT:  while inserting index tuple (8304,21) in relation "residual"
STATEMENT:  
             INSERT INTO residual (residual)
             SELECT DISTINCT residual FROM measurement_blob_
             ON CONFLICT DO NOTHING

```
</details>

- [x] DNS ttl is uint32 and some quite interesting replies with TLL 3222011910 and 4294967292...4294967295 do not fit into int32
```
ERROR:  value "4294967292" is out of range for type integer
CONTEXT:  COPY dns_a_, line 5543, column ttl: "4294967292"
STATEMENT:  COPY dns_a_(msm_no,domain,control_ip,test_ip,control_cname,test_cname,ttl,reso...
```

- [x] `status_code: "2725348804"` in http_request
```
ERROR:  value "2725348804" is out of range for type smallint
CONTEXT:  COPY http_request, line 3145, column status_code: "2725348804"
STATEMENT:  COPY http_request(msm_no,url,failure,status_code,body_length,tit
```

- [ ] `report_id` is not unique, there are constraint violations

<details>

```
ERROR:  duplicate key value violates unique constraint "report_report_id_key"
DETAIL:  Key (report_id)=(cM3Xib672fUG4CYtgdScJ9TgvDDWpuh88pN9a1GsPDGHtRM3COgJg03bzXCXpwgE) already exists.
STATEMENT:  
                 INSERT INTO report
                 SELECT report_no, autoclaved_no, test_start_time, probe_cc, probe_asn, probe_ip,
                        test_name, badtail, textname, orig_sha1,
                        COALESCE(report_id, translate(encode(orig_sha1, 'base64'), '/+=', '_-')), software_no
                 FROM report_meta_
                 JOIN report_blob_ USING (report_no)
```
and
```
ERROR:  duplicate key value violates unique constraint "report_report_id_key"
DETAIL:  Key (report_id)=(20150901T130829Z_SZvCaHrEjwRncwjCelzGHbxMUzNmvQCNrfFFdZsGyBOGKgCpLs) already exists.
STATEMENT:  
                 INSERT INTO report
                 SELECT report_no, autoclaved_no, test_start_time, probe_cc, probe_asn, probe_ip,
                        test_name, badtail, textname, orig_sha1,
                        COALESCE(report_id, translate(encode(orig_sha1, 'base64'), '/+=', '_-')), software_no
                 FROM report_meta_
                 JOIN report_blob_ USING (report_no)
```
</details>

- [ ] sometimes DNS domain is mistakenly taken as `null` 
```
ERROR:  null value in column "domain" violates not-null constraint
DETAIL:  Failing row contains (11293458, null, null, null, null, null, null, null, null, null, deferred_timeout_error).
CONTEXT:  COPY dns_a_, line 5956: "11293458      \N      \N      \N      \N      \N      \N      \N      \N      \N      deferred_timeout_error"
STATEMENT:  COPY dns_a_(msm_no,domain,control_ip,test_ip,control_cname,test_cname,ttl,resolver_hostname,client_resolver,control_failure,test_failure) FR
```

- [x] `\u0000` replacing is not properly working for some reason
```
ERROR:  unsupported Unicode escape sequence
DETAIL:  \u0000 cannot be converted to text.
CONTEXT:  JSON data, line 1: ...net (ApacheTrafficServer [cMsSf ])","Set-Cookie":...
 COPY http_control, line 3706, column headers: "{"Via":"https\/1.1 r14.ycpi.dcb.yahoo.net (ApacheTrafficServer [cMsSf ]), https\/1.1 r07.ycpi.loa.ya..."
STATEMENT:  COPY http_control(msm_no,is_tor,failure,status_code,body_length,title,headers,body_sha256) FROM stdin WITH DELIMITER AS '    ' NULL AS '\N'
```

- [ ] control_failure `[ambiguous]` is stamped at ~27% of `dns_a` rows, that's likely a bug
```
metadb=# create temporary table dns_sample as select * from dns_a tablesample system (1);
metadb=# select control_failure, (count(*) * 100. / (select count(*) from dns_sample))::numeric(5,2) AS pct from dns_sample group by control_failure;
  control_failure   |  pct  
--------------------+-------
                    | 72.34
 [ambiguous]        | 27.65
 dns_name_error     |  0.00
 dns_server_failure |  0.01
```

- [ ] control_ip is not null only in ~4% or `dns_a` rows,  test_ip is not null only for ~24% of rows, 76% of rows have both control_ip and test_ip being null, looks like bug too.

- [ ] whole bucket is skipped in case of TTL/report_id errors as it's unclear how to degrade cleanly, maybe [rollback to savepoint](https://www.postgresql.org/docs/current/static/sql-rollback-to.html)? is it fast enough? how should the bucket be marked as dirty in this case?

darkk:
Current checklist to stop `luigi_pipeline`:

MVP:

- [ ] PG @ HKG
  - [ ] Setup a dedicated read-only user for ooni-measurements
  - [ ] Expose postgres database from HKG
- [ ] centrifugation updates
  - [x] check if detected deadlock leads to skipped bucket or not as simple retry may be enough right now
  - [x] signature-based incremental blockpage detection
  - [x] sensor checking if another bucket is cut off at chameleon (both pipelines are active during transition period) and airflow-pipline can do another 24h tick
  - [x] write to a <s>single</s> log file enough details to debug issue of malformed data in line `copy_from` blob (it should include <s>raw report file, offset into it and stack trace</s> enough details)
  - [ ] check for a bug leading to `dns_a` table having 75% of "empty" rows

Next steps:

- [ ] PG @ AMS
  - [ ] streaming replication from HKG to AMS
  - [ ] securing replication with TLS for integrity reasons
  - [ ] read-only user for replica
  - [ ] moving the postgre replica to SSD
- [ ] data proxy @ AMS
  - [ ] deploy & test nginx caching proxy in AMS that can handle partial request and maintain keep-alive to HKG to keep CWND while tcp_slow_start_after_idle=0
  - [ ] auto-cleanup procedure for nginx to maintain cache size
  - [ ] use varnish and squid as a fallback if nginx fails (AFAIR, it had all required features)
- [ ] centrifugation updates
  - [ ] do ROLLBACK TO SAVEPOINT in case of malformed data line in `copy_from` blob & mark bucket as dirty
  - [ ] check for possible table size reduction methods for `dns_a` (29 Gb, 35%), `http_request` (24 Gb, 29%) and `http_control` (16 Gb, 20%) -- don't implement them at the moment, just understand the stats
- [ ] add aforementioned network services to blackbox_exporter monitoring

Everything that is required to shutdown chameleon altogether (after transition of services from `ooni_pipeline` DB to metadata DB) is not mentioned in this checklist: moving data from collectors to pipeline, s3, backups, cleanups, etc.

Questions:
- Do we want to setup ACL for data proxy (if so maybe post MVP)?

Edit: I made some minor changes to separate what is MVP and what is not ~ A.

darkk:
> postgres detects deadlock on three concurrent writes to `input` and `residual` tables

There were 45 "deadlock detected" cases, 6 of them observed another deadlock during retry. Additional retry produces some sane output. I don't care about deadlocks for a while anymore.

> DNS ttl is uint32 and some quite interesting replies with TLL 3222011910 and 4294967292...4294967295 do not fit into int32

It's anomaly, RFC states that TTL is positive signed int31. Negative values raise exception now.

> status_code: "2725348804" in http_request

I assume it's just a bug in MK (TheTorProject/ooniprobe-android#93), bad values raise exception now.


Also, I'd like to record some numbers to support backfilling distributed across several VMs when `FLAG_SKIP_SIMHASH` will be turned off:
- iperf3 between to VMs in HKG produces ~5..6 Gbit/s that's 625+ Mbyte/s
- single-threaded sequential disk read at datacollector.infra.ooni.io produces ~58 Mbyte/s
- six-threaded sequential disk read at datacollector.infra.ooni.io produces ~200 Mbyte/s
- ~same disk performance is observed via NFS
- `sim_shi4_mm3` computation consumes ~7.5 Mbyte/s
- ⇒ it makes _some_ sense to use up to ~6 VMs having 4 cores each, postgresql sink performance is not taken into account.

- [x] check if every fingerprint is triggered at least once

darkk:
> check if every fingerprint is triggered at least once
```
179003	1	IR	iframe src="http://10.10		
1040	2	TR	<title>Telekomünikasyon İletişim Başkanlığı</title>		
10525	3	GR	www.gamingcommission.gov.gr/index.php/forbidden-access-black-list/		
2781	4	RU	http://eais.rkn.gov.ru/		
578	5	IN	The page you have requested has been blocked		
1085	6	BE	that is considered illegal according to Belgian legislation		
751	7	IT	GdF Stop Page		
4	8	CY	nba.com.cy/Eas/eas.nsf/All/6F7F17A7790A55C8C2257B130055C86F		
73	9	DK	lagt at blokere for adgang til siden.		
1486	10	FR	xtpage = "page-blocage-terrorisme"		
15553	11	MY	Makluman/Notification		
24813	12	KR	http://warning.or.kr		
295	13	RO	Accesul dumneavoastră către acest site a fost restricționat		
1486	14	GF	xtpage = "page-blocage-terrorisme"		
3	15	ID		Location	http://internet-positif.org
323	16	UK		Location	http://blocked.nb.sky.com
633	17	RU		Location	http://warning.rt.ru
12022	18	SA		Server	Protected by WireFilter
12022	19	AE		Server	Protected by WireFilter
10	20	SD		Location	http://196.1.211.6:8080/alert/
3	21	QA		Location	http://www.vodafone.qa/alu.cfm
3604	22	KR		Location	http://www.warning.or.kr
7	23	PT		Location	http://mobilegen.vodafone.pt/denied/dn
182	24	NO		Location	http://block-no.altibox.net/
```

af/shovel/check_sanitised.py
```diff
@@ -0,0 +1,163 @@
...
+import hashlib
+import os
+import tempfile
+from functools import partial
+
+import ujson
+
+import autoclaving
+from canning import dirname
+from centrifugation import stream_datum, FILE_START, FILE_END, REPORT_START, REPORT_END, BADBLOB, DATUM
```
hellais: `stream_datum` is not defined in centrifugation (anymore?) and is causing the `sanitised_check` task to fail in the `hist_canning` dag.

see:

```
[2017-08-24 00:23:10,171] {base_task_runner.py:95} INFO - Subtask: [2017-08-24 00:23:10,170] {bash_operator.py:94} INFO - File "/usr/local/bin/check_sanitised.py", line 21, in <module>
[2017-08-24 00:23:10,171] {base_task_runner.py:95} INFO - Subtask: [2017-08-24 00:23:10,170] {bash_operator.py:94} INFO - from centrifugation import stream_datum, FILE_START, FILE_END, REPORT_START, REPORT_END, BADBLOB, DATUM
[2017-08-24 00:23:10,171] {base_task_runner.py:95} INFO - Subtask: [2017-08-24 00:23:10,171] {bash_operator.py:94} INFO - ImportError: cannot import name stream_datum
[2017-08-24 00:23:11,888] {base_task_runner.py:95} INFO - Subtask: [2017-08-24 00:23:11,888] {bash_operator.py:97} INFO - Command exited with return code 1
```

hellais (approved):
I think this is good to land into master. Let's create issues for all the stuff that has not yet been done and merge it 🐳

darkk:
Yep, there is some technical debt to be resolved, but current pipeline is already in a better and more stable state then luigid-based one in terms of error handling, reliability and speed and supported capacity.
These pipelines are going to coexist for a while minor mismatches in **sanitized** data are investigated and explorer is migrated to new API, although luigid should considered frozen legacy pipeline.
Cleanup of luigid code will happen after #94 converges.

PR #71, opened 2017-07-27
Always compress "compressible" files in S3

fortuna:
I'm starting this PR for discussion.

The idea here is to always store files in S3 in compressed format. The OONI data is highly compressible, so this should significantly reduce the costs for storage and transfer, as well as make the transfers a lot faster.

This change should be backward compatible: people accessing the files from S3 will only get compressed content if they set `Accept-Encoding: gzip`.

Note that I have not tested this change. I don't believe I can easily, so I'd love some guidance. It may make more sense for someone from the OONI team to test it instead.

### DIFF
--- a/pipeline/batch/daily_workflow.py
+++ b/pipeline/batch/daily_workflow.py
@@ -197,6 +197,10 @@ def normalise_str(body):
             body = binary_to_base64_dict(body)
     return body
 
+def file_is_compressible(filename):
+    """Returns true for file types that benefit from compression."""
+    return filename.endswith('.json') or \
+           filename.endswith('.yaml')
 
 def get_luigi_target(path):
     try:
@@ -211,9 +215,14 @@ def get_luigi_target(path):
 
     file_format = None
     if path.endswith(".gz"):
-        file_format = GzipFormat()
+        file_format = GzipFormat(compression_level=9)
     if path.startswith("s3n://"):
-        return S3Target(path, format=file_format)
+        # Try to always store files in S3 in compressed format.
+        s3_headers = {}
+        if file_is_compressible(path):
+            file_format = GzipFormat(compression_level=9)
+            s3_headers["Content-Encoding"] = "gzip"
+        return S3Target(path, format=file_format, headers=s3_headers)
     elif path.startswith("ssh://"):
         ssh_key_file = config.get("ssh", "ssh-key-file", None)
         no_host_key_check = config.get("ssh", "no-host-key-check", None)


darkk:
> I'm starting this PR for discussion.

Current plan regarding storage is to store "large" reports compressed with lz4 and small reports -- with tar.lz4 where lz4 frames are aligned with reports inside of tar file to make tar file "seekable". That's already implemented in pipeline-16.10 branch, but it's not part of public API yet: two versions of pipeline are currently running side-by-side.

lz4 was chosen instead of gzip because of comparable compression ratio (IIRC, 22% vs 19%) and much faster decompressor.

> This change should be backward compatible

Thanks for valuable information. I assumed that S3 can't provide transparent gzip decompression to old clients.

> reduce the costs for storage and transfer

Current plan is to close access to S3 gracefully and abuse generosity of our hosting provider to serve the data from another location that costs us much less. There is beta API to fetch single measurements and there can be API for to fetch compressed json.lz4 / tar.lz4 files.

darkk:
> Current plan is to close access to S3

We've applied to Amazon Open Data program and have got some "free" storage & transfer. Note, it does not apply to `s3://ooni-public` bucket that will likely be unpublished soon.

> tar.lz4 .. is not part of public API yet

These tar.lz4 files are now available at `s3://ooni-data/autoclaved/jsonl.tar.lz4/`. They [should be readable](https://github.com/TheTorProject/ooni-pipeline/blob/84e738e71ac65cdc62cb816c91dc0e6d776b6448/af/shovel/aws_s3_lz4cat_sync.py#L119-L126) with ordinary `tar -I lz4 --extract` from ubuntu:16.04. It's not a part of "official" public API yet, but it's here to stay for some time as this format solves some efficiency issues of underlying `jsonl` format.

Uncompressed data replacing `s3://ooni-public/sanitised` is available as `s3://ooni-public/autoclaved/jsonl`. It's renamed as it's generated by a slightly different code and may have slightly different binary representation of old files. I mean stuff like different random-generated `id`, different float representation, field order and so on.

darkk:
I think, that can be closed.

That ugly `*.tar.lz4` is described in blog post: https://ooni.torproject.org/post/mining-ooni-data/

The main reason to have it was to reduce ~6M report files taking ~18TiB to ~0.2M archives taking ~3.8 TiB  to make it more friendly to HDD-based storage and avoid IO operations on reports as small as ~4kb, while preserving compatibility with possible existing consumers.

Other issues of current data format are described at #114 quite well.

PR #97, opened 2017-10-16
Docs/3.0.0

hellais:
In this PR I update the Readme with usage and information about the new pipeline.


Readme.md
```diff
@@ -0,0 +1,243 @@
+# OONI pipeline
+
+version: 3.0.0
+
+> If OONI were a car, the OONI Pipeline would be it's engine
```
darkk, line 5: > If Tor is an Guy Fawkes'es umbrella and OONI is octopus-shaped network investigator under that umbrella, the OONI Pipeline is the steam engine of this motorized octopus!

Readme.md
```diff
@@ -0,0 +1,243 @@
...
+ not seconds.
+
+* Make it easy to re-run new heuristics on top of the existing dataset
+
+* Ease parallelization of CPU intensive tasks to use all available cores on a
+ single machine
+
+## Architecture
+
+![Pipeline Architecture](.assets/PipelineArchitecture.png)
```
darkk, line 23: **If** you want to make that editable, please, use `dot` (or something having similar properties).
It's ugly, but it's so easy to update and extend it :)

Readme.md
```diff
@@ -0,0 +1,243 @@
...
+  | canning-2.py | lz4 -5
+  | canning-3.py
+```
+
+`canning-1.py`: groups reports by `test_name`, slices them into ~64 Mbytes groups of
+uncompressed data and send them to `tar`.
+
+`canning-2.py`: reads a `.tar` file, verifies the sizes of reports, calculates
+the sha1 and crc32 of each report, calculate size of tar file and then pipes it
+to `lz4`.
```
darkk, line 50:
```diff
-and then pipes it
+and then pipes the unmodified tar file
```

Readme.md
```diff
@@ -0,0 +1,243 @@
...
+lz4 compression is a bit tricky at this stage: lz4 frames are independent,
+report offsets within the block are recorded in the separate index file, so the
+resulting file is both streamable (with ability to skip tar records without
+decoding), seekable (single lz4 frame is read to get a single report — that's
+~56kb of compressed data) and readable with ordinary `tar -I lz4 --extract` for
+whoever wants to parse sanitized measurements.
+
+Indexes for the reports are stored as well, they include block offsets, sizes,
+and offsets within block required to seek() to exact measurement and to read the
+file in streaming way without parsing tarfile stream.
```
darkk, line 79: > without parsing tarfile stream.

That's not actually a goal, but an implementation detail. Streaming with `tarfile.py` is kinda slow, that's true, but another way to achieve same goal with similar performance is just to read `tar -I lz4 --to-stdout`. I'm unsure if there was significant performance difference between these two way of streaming. The reason to chose that implementation instead of `tar -I lz4` subprocess was to consume CPU within single process to ease parallel computing "accounting".

Readme.md
```diff
@@ -0,0 +1,243 @@
...
+
+The file to edit is located in `af/shovel/centrifugation.py`.
+
+2. Push a new docker image
+
+You should then build a new docker image by bumping the version number inside of
+`af/shovel/build`.
+
+3. Redeploy an update
+
```
darkk, line 135: Version number in `ansible/roles/airflow/files/docker-trampoline` should be bumped as well.

Readme.md
```diff
@@ -0,0 +1,243 @@
...
+times.
+
+1. Edit the centrifugation file
+
+The file to edit is located in `af/shovel/centrifugation.py`.
+
+2. Push a new docker image
+
+You should then build a new docker image by bumping the version number inside of
+`af/shovel/build`.
```
darkk, line 132: And push built image to dockerhub. `docker push openobservatory/shovel:latest && docker push openobservatory/shovel:0.0.NN`

Readme.md
```diff
@@ -0,0 +1,243 @@
...
+    ''
+) from fingerprint
+```
+
+5. Deploy and re-process the data
+
+Basically you need to do the same thing you did in "Modifying the centrifugation
+stage" from step 2. onwards.
+
+The task you want to re-run is labelled `meta_pg`.
```
darkk, line 218:
```diff
+Reprocessing of whole dataset takes couple of days, it's done asynchronously,
+but the pipeline does not implement any priorities, so it may block data ingestion for a while.
```

Readme.md
```diff
@@ -0,0 +1,243 @@
...
+## To extend the DAG
+
+You need to edit the DAG inside of ooni-sysadmin repo:
+https://github.com/TheTorProject/ooni-sysadmin/tree/master/ansible/roles/airflow/files/airflow-dags
+
+Add a `BashOperator`, get the `task_id`.
+
+Then add a new switch case inside of
+`ansible/roles/airflow/files/docker-trampoline`.
+
```
darkk, line 229: IMHO, it deserves an example when the DAG should be extended and when another DAG should be created.
- extend: fetch data from collectors or backup stuff to S3 after processing
- new dag: fetch data regarding ASNs (as there is no dependency on other tasks)

Also, DAGs may trigger another DAGs, but DAGs are "pausable" so it's another way to reason about extending vs. creation. E.g. one may want to be able to pause processing, but continue ingestion. But I prefer to avoid about thinking about that aspect till we have a concrete case.

af/.gitignore
```diff
@@ -0,0 +1,6 @@
+# TMP
+airflow/dumb-init_1.2.0_amd64.deb
+shovel/*.flame
+shovel/*.svg
+shovel/bridge_db.json
+useless-test-dags/
```
darkk, line 6: Were these files ever commited?...

Readme.md
hellais: If we need to change this I have the editable source of the file and can take care of making the changes in it myself.

af/.gitignore
hellais: Don't think so, but I had them lying around on my filesystem so I populated the gitignore with them

Readme.md
hellais: Should I in some way integrate this comment into the docs? If so maybe you can make this edit as I am not really sure where it should go.

Readme.md
darkk: > I have the editable source of the file

SPOF! SPOF! :-D

Yeah, that's not important, I just stated that once again: `dot` and `seqdiag` are great tools to make diagrams when their expressiveness is enough.

Readme.md
darkk: Okay

Readme.md
hellais: I like pretty charts 😄

PR #98, opened 2017-10-17
Feature/tor metrics

hellais:
Adds support for extracting features from `vanilla_tor` measurements.

This depends on https://github.com/TheTorProject/ooni-pipeline/pull/97

### DIFF
--- a/af/oometa/006-vanilla-tor.install.sql
+++ b/af/oometa/006-vanilla-tor.install.sql
@@ -8,7 +8,7 @@ select _v.register_patch( '006-vanilla-tor', ARRAY[ '005-repeated-report' ], NUL
 CREATE TABLE vanilla_tor (
     msm_no                  int4 references measurement,
     success                 bool null,
-    error                   bool null,
+    error                   text,
     timeout                 int4,
     tor_progress            int4,
     tor_progress_tag        text,
--- a/af/oometa/006-vanilla-tor.install.sql
+++ b/af/oometa/006-vanilla-tor.install.sql
@@ -4,18 +4,18 @@ select _v.register_patch( '006-vanilla-tor', ARRAY[ '005-repeated-report' ], NUL
 
 -- Everything goes to `public` schema.
 
-
 CREATE TABLE vanilla_tor (
-    msm_no                  int4 references measurement,
-    success                 bool null,
+    msm_no                  int4 not null, -- references measurement,
+    timeout                 int4 not null,
     error                   text,
-    timeout                 int4,
-    tor_progress            int4,
+    tor_progress            int2 not null,
+    success                 bool not null,
     tor_progress_tag        text,
     tor_progress_summary    text,
-    tor_version             text,
-    tor_log                 text,
-    transport_name          text
+    tor_version             text not null,
+    tor_log                 jsonb
 );
+
+COMMENT ON TABLE vanilla_tor IS 'Features: data from `vanilla_tor` measurements';
 
 COMMIT;
--- a/af/oometa/006-vanilla-tor.install.sql
+++ b/af/oometa/006-vanilla-tor.install.sql
@@ -1,0 +1,21 @@
+BEGIN;
+
+select _v.register_patch( '006-vanilla-tor', ARRAY[ '005-repeated-report' ], NULL );
+
+-- Everything goes to `public` schema.
+
+
+CREATE TABLE vanilla_tor (
+    msm_no                  int4 references measurement,
+    success                 bool null,
+    error                   bool null,
+    timeout                 int4,
+    tor_progress            int4,
+    tor_progress_tag        text,
+    tor_progress_summary    text,
+    tor_version             text,
+    tor_log                 text,
+    transport_name          text
+);
+
+COMMIT;
--- a/af/shovel/centrifugation.py
+++ b/af/shovel/centrifugation.py
@@ -26,6 +26,7 @@ import ujson
 
 import autoclaving
 from canning import isomidnight, dirname
+from tor_log import parse_tor_log
 
 # It does NOT take into account metadata tables right now:
 # - autoclaved: it's not obvious if anything can be updated there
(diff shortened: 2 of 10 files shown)


af/oometa/006-vanilla-tor.install.sql
```diff
@@ -0,0 +1,21 @@
+BEGIN;
+
+select _v.register_patch( '006-vanilla-tor', ARRAY[ '005-repeated-report' ], NULL );
+
+-- Everything goes to `public` schema.
+
+
+CREATE TABLE vanilla_tor (
+    msm_no                  int4 references measurement,
```
darkk: `references` is slow as it's a trigger that is running on every insertion (and I have not found an easy way to disable constraints check, do it in batch and re-enabled later), it should be plain `integer NOT NULL`.

If you want to have constraint check, it should be done differently in our case, my proposal is to define a function like ooni_constraints() that checks all the required cross-table assertions.

af/oometa/006-vanilla-tor.install.sql
```diff
@@ -0,0 +1,21 @@
...
+    success                 bool null,
+    error                   bool null,
+    timeout                 int4,
+    tor_progress            int4,
+    tor_progress_tag        text,
+    tor_progress_summary    text,
+    tor_version             text,
+    tor_log                 text,
+    transport_name          text
```
darkk: Actually, everything that is not `NOT NULL` is nullable by default :)
If there are some values that are never `NULL`, mark them as `NOT NULL` to save some bits and have cheap DB-level sanity check.

darkk (commented):
I'll push further changes directly to this branch.

darkk:
> https://github.com/TheTorProject/ooni-pipeline/pull/98/commits/da2fd0aed218f5c5912f6d5c4aa9f64c1d538b89 centrifugation: split steps `ingest`, `reingest` and `reprocess`

Single-point estimate shows 3.5% performance degradation. It's good enough, so I don't care to measure precisely :)

af/shovel/tor_log.py
```diff
@@ -0,0 +1,95 @@
...
+
+        if msg in MSG_BOOTSTRAP:
+            ret.append({'t': ts, '%': MSG_BOOTSTRAP[msg]})
+        else:
+            ret.append({'t': ts, '_': msg})
+    return ret
+
+
+if __name__ == '__main__':
+    def main():
```
hellais, line 86: Why do you double wrap the `main()`?

Is it to avoid exporting this function? If that is the case maybe it's cleaner to put it before the `if __name__ == '__main__'` and call it `_main()` to make it clear it's a private function

af/shovel/tor_log.py
```diff
@@ -0,0 +1,95 @@
...
+
+        if msg in MSG_IGNORE or any(r.match(msg) for r in MSG_RE_IGNORE):
+            continue
+
+        if date == base_date:
+            ts = ts - base_ts
+        elif date[1] == base_date[1] + 1 or date[1] == 1:
+            ts = ts - base_ts + 86400
+        else:
+            raise RuntimeError('tor_log spans over two+ days', base_date, date)
```
hellais, line 75: Maybe if this happens we should not throw an error, but rather mark it as time warping and still ingest it?

af/shovel/centrifugation.py
```diff
@@ -326,163 +329,491 @@ def __flush_stubborn(self):
...
-
-    files = set()
-    use_report = True
-    report_count, datum_count = 0, 0
+def load_autoclaved_index(autoclaved_index):
+    # Returns: {filename: file_sha1 for autoclaved}
+    files = {}
     with gzip.GzipFile(autoclaved_index, 'r') as indexfd:
         for _, doc in autoclaving.stream_json_blobs(indexfd):
+            if 'file"' not in doc:
```
hellais: I would maybe add a comment above this stating that this is a "fast path" and that the `"` after file is not a typo, but rather a slightly more strict check for the JSON key `"file": something` (also why not `file":`?)

af/shovel/centrifugation.py
```diff
@@ -326,163 +329,491 @@ def __flush_stubborn(self):
...
-        elif len(files_db) == 0: # and len(files) > 0 as sets are not equal :)
-            return None
+        c.execute('SELECT filename, file_sha1, code_ver FROM autoclaved WHERE bucket_date = %s', (bucket,))
+        for filename, file_sha1, code_ver in c:
+            file_sha1 = str(file_sha1) # file_sha1 is returned as buffer()
+            if filename not in files:
+                raise RuntimeError('Unknown filename in autoclaved PG table present in autoclaved/index.json.gz', filename)
+            elif files[filename] != file_sha1: # autoclaved file changed, e.g. PII cleanup
+                reingest.add(filename)
+            elif code_ver < CODE_VER: # including CODE_VER_REPROCESS
```
hellais: I don't understand the comment here. Should this comparison be against `CODE_VER_REPROCESS` or include `CODE_VER_REPROCESS` in some way?

af/shovel/centrifugation.py
```diff
@@ -35,6 +36,8 @@
 # causing bucket-wide exception, so re-import is not forced.
 CODE_VER = 4
 
+CODE_VER_REPROCESS = 0 # special `code_ver` value to do partial re-processing
```
hellais: Can you expand this comment explaining the use case of `CODE_VER_REPROCESS`? Documentation on it's usage should also be added to the README docs.

af/shovel/centrifugation.py
```diff
@@ -326,163 +329,491 @@ def __flush_stubborn(self):
...
+                reprocess.add(filename)
+            elif code_ver == CODE_VER:
+                pass # no-op
+            else: # including code_ver > CODE_VER
+                raise RuntimeError('Bad autoclaved row', filename, file_sha1, code_ver)
+            files.pop(filename)
+    ingest = files.viewkeys() # files that are not in the database at all
+    return ingest, reingest, reprocess
+
+def create_tmp_table(pgconn, table, definition):
```
hellais: <bikeshed>
For consistency maybe this should be called `create_temp_table`
</bikeshed>

af/shovel/centrifugation.py
```diff
@@ -326,163 +329,491 @@ def __flush_stubborn(self):
...
+def copy_meta_from_index(pgconn, ingest, reingest, autoclaved_index, bucket):
+    # Writing directly to `report` table is impossible as part of `report`
+    # table is parsed from data files, same is true for `measurement` table.
+    # The reason for `autoclaved` is cross-reference generation, it's possible
+    # to avoid it, but the price is useless extra complexity. Filling these
+    # three tables one-by-one is also possible, but it makes re-ingestion
+    # preserving `*_no` sequence values and available range a more complex task
+    # as sequence is non-transactional.
+    if not DUPLICATE_REPORTS:
+        # FIXME: eventually this assertion should be removed and DUPLICATE_REPORTS
```
hellais: I am not sure I agree with this. There are cases in which duplicate reports can also span multiple buckets (for example a user did an upload of a measurement they already submitted days after).

That said we should probably be detecting duplicates much earlier (maybe in the collector itself) and reject them there.

af/shovel/centrifugation.py
```diff
@@ -635,37 +1027,52 @@ def exc_hash(exc_info):
     ret |= mmh3.hash(msg) & 0xffff
     return to_signed32(ret)
 
-class BadmetaFeeder(object):
-    table = 'badmeta_' # it's actually some temporary table
-    columns = ('autoclaved_no', 'report_no', 'textname', 'exc_report', 'exc_measurement')
+class BadmetaFeeder(BaseFeeder):
+    sink_table = 'badmeta_'
```
hellais: Is this no longer a temporary table?

docs/pipeline-16.10.md
```diff
@@ -170,6 +170,27 @@ Highlights
...
+
+Reingestion is usually caused by PII cleanup or other updates to *autoclaving*
+step. Reingestion is usually trigged by changed `file_sha1` of autoclaved file.
+
+Reprocessing is triggered by `code_ver`. Every autoclaved file has
+`autoclaved.code_ver` mark in the metadata postgres DB and whole step has
+`centrifugation.CODE_VER` as well set in `centrifugation.py`.
+If `code_ver` and `CODE_VER` match, the file is not processed at all, it's not
+even read. If `code_ver` does not match, then the file is read, but it depends
+on `code_ver` column and `compat_code_ver` attribute if corresponding tables
```
hellais: Should this not say `min_compat_code_ver`?

af/shovel/centrifugation.py
```diff
@@ -326,163 +329,491 @@ def __flush_stubborn(self):
...
+            elif files[filename] != file_sha1: # autoclaved file changed, e.g. PII cleanup
+                reingest.add(filename)
+            elif code_ver < CODE_VER: # including CODE_VER_REPROCESS
+                reprocess.add(filename)
+            elif code_ver == CODE_VER:
+                pass # no-op
+            else: # including code_ver > CODE_VER
+                raise RuntimeError('Bad autoclaved row', filename, file_sha1, code_ver)
+            files.pop(filename)
+    ingest = files.viewkeys() # files that are not in the database at all
```
hellais, line 372: Attention, this is a python2-ism

af/shovel/centrifugation.py
```diff
@@ -326,163 +329,491 @@ def __flush_stubborn(self):
...
-                use_report = True
-            elif t == 'datum' and use_report:
-                datum_count += 1
-
+            elif doc['type'] == '/file':
+                files[filename] = b64decode(doc['file_sha1'])
+    return files
+
+def load_autoclaved_db_todo(pgconn, files, bucket):
+    # Returns: {files to ingest}, {files to re-ingest}, {files to re-process}
```
hellais, line 351: Maybe it's wise to mention that the returned type is a `dict_keys()` instance and two `set()`s

af/shovel/centrifugation.py
```diff
@@ -326,163 +329,491 @@ def __flush_stubborn(self):
...
-    columns = ('report_no', 'frame_off', 'frame_size', 'intra_off', 'intra_size', 'orig_sha1')
-    with closing(PGCopyFrom(pgconn, table, columns=columns)) as wbuf, \
-        gzip.GzipFile(autoclaved_index, 'r') as indexfd:
-        rowno = None
-        # Separate `use_report` variable is used instead of `rowno=None` to
-        # enforce implicit file format checking.
-        use_report = True
+            raise AssertionError('BUG: should never happen')
+        dest_wbuf[(doc_type, key)].write(copy_row)
+    with gzip.GzipFile(autoclaved_index, 'r') as indexfd:
```
hellais, line 456: This whole function is very long, maybe  it could be refactored into a smaller function (the switch case inside of the for loop) for readability.

af/shovel/centrifugation.py
```diff
@@ -326,163 +329,491 @@ def __flush_stubborn(self):
...
+                        doc['text_off'] - frm['text_off'], # intra-frame off, as text_off is absolute
+                        doc['text_size'], # intra-frame size equals to stream size :)
+                        pg_binquote(b64decode(doc['orig_sha1']))))
+            else:
+                raise RuntimeError('Unknown index.json.gz row type', autoclaved_index, doc)
+    # `dest_wbuf` is not managed with closing() to avoid masking one exception
+    # with another in case of failure.
+    for wbuf in dest_wbuf.itervalues():
+        wbuf.close()
+    with pgconn.cursor() as c:
```
hellais: This should probably be wrapped in a transaction and probably also placed inside of a separate function that takes the argument `bucket`

af/shovel/centrifugation.py
darkk: > why not `file":` ?

because it's `"type": "file"` and `"type":"/file"` :)

af/shovel/centrifugation.py
darkk: Well, I consider that it's mentioned. `{1, 2, 3}` is syntax for `set()`. `dict_keys()` is set-like object, so I don't know if it's possible to distinguish it from set using usual set methods without type switch.

af/shovel/centrifugation.py
darkk: CODE_VER_REPROCESS = 0, it's less than any `CODE_VER` that ever existed.

af/shovel/centrifugation.py
darkk: True, there is another one somewhere else (IIRC, around crc32). What's py3-way? `set(file.keys())` ?

af/shovel/centrifugation.py
darkk: Consistency... with what? :) What do you mean as a reference point? There are points for both cases: `tempfile`, `TMPDIR`, `TMP_MAX`, `mktemp`...

af/shovel/centrifugation.py
darkk: True. I meant that global `DUPLICATE_REPORTS` _does not scale_ :-)
Doing that at highly available collector is also not 100% trivial, but collector may easily do some pre-filtering as well. Anyway, it's not a big deal at **this** moment.

af/shovel/centrifugation.py
darkk: I tried. And failed. There is shared context between all cases under the switch, that's why extracting handling to separate functions does not improve readability that much.
So I decided to keep it in that shape for a while.

af/shovel/centrifugation.py
darkk: It is, but there is no reason to state it explicitly anymore. `table` attr was EITHER persistent table OR temporary one and it was ugly and misleading and error-prone and bad and weed and Monty Python disapproved that decision. `table` is split into `data_table` (persistent) and `sink_table` (scratchpad to write rows to).
There is no corresponding `data_table` as `badmeta` is not used as a `DATA_TABLE`.

af/shovel/tor_log.py
darkk: The measurement itself will be ingested, but there will be no corresponding row in `vanilla_tor` table. I think, it's okay to throw as that outlier is "special" and requires special handling and special attention. I think, it's okay to gather these special cases marked with some exception in `exc` column and check them later (as this _process_ is also something that should be done).

af/shovel/tor_log.py
darkk: > Is it to avoid exporting this function

True. It's not _private_ code, it's sort of _debugging_ code. That's the same hack some stdlib modules use :)

af/shovel/centrifugation.py
hellais: Ah ok I see, so you are actually looking for the JSON "value", not the JSON key. I think it's worthwhile to add a comment saying what it is that you are looking for then.

af/shovel/centrifugation.py
darkk: Actually, the comment is there to make code grepable for "CODE_VER_REPROCESS" string :)

af/shovel/centrifugation.py
darkk: It's actually there, but deserves some more words:

```diff
+Both these cases are handled with `UPDATE autoclaved SET code_ver = 0 WHERE …`.
+`centrifugation.py` will re-ingest alike autoclaved file while preserving `msm_no`.
```

af/shovel/centrifugation.py
hellais: I mean in this file, I only see `FLAG_TRUE_TEMP` (which is talking about a temporary table).

Also in the python standard library the tradition is to call temporary related function `temp`, example: `tempfile` `mktemp` etc.

af/shovel/centrifugation.py
hellais: > What's py3-way? 

`dict.keys()` ;)

hellais: Also if we want to eventually upgrade to python3, I would suggest we start using something like `six` to take care of these situations.

af/shovel/centrifugation.py
hellais: Fair enough. I would still prefer to explicitly state it as I was unsure in the beginning if you were using that syntax to say they are lists or sets.

af/shovel/centrifugation.py
hellais: ACK

af/shovel/tor_log.py
hellais: Ok, fair enough.

af/shovel/centrifugation.py
darkk: > This should probably be wrapped in a transaction

`meta_pg()` defines "large" transaction in `with pgconn, ...: # main transaction`.

af/shovel/centrifugation.py
darkk: `TMP_MAX` is also from stdlib, from `tempfile` module. ;-)
But, okay, I'll rewrite the history a bit if you prefer that name. And, obviously, you do, as you're motivated enough to write at least two comments to change single char :-D

darkk:
Works as expected, further improvements will follow as separate branches.

PR #101, opened 2017-10-25
Add aws_s3_cp.py command

hellais:


af/shovel/aws_s3_lz4cat_sync.py
```diff
@@ -0,0 +1,64 @@
...
+            continue
+        # Decompressing files chunk-by-chunk saves page cache from pollution.
+        # It's possible to stream files from local lz4 to S3 without touching
+        # the disk and it *seems* to be easy to do that, as 80% of the volume
+        # consists of "large" reports, but it makes usage of `aws s3 sync`
+        # impossible, it would force me to write my own syncing code for
+        # idempotent upload without wasted bandwidth and it's not something I
+        # want to do right now. It may be reconsidered if we want to preserve
+        # YYYY-MM-DD/xxxx naming scheme while transitioning to hourly
+        # batches.  -- @darkk 2017-11-10
```
darkk, line 42: Well... it's slow as hell (time is *hours*):
airflow-jsonl-n2 2017-11-11 13-23-25

darkk: That's one bug contributing to slowness: https://github.com/TheTorProject/ooni-pipeline/commit/fd86c9cf45918aef69b8bd49ca436260a09b1827
Another issue is that `aws s3 sync` probably scans whole tree even if there is no `--delete`, so amount of fetched data is ~ O( lz4_files_that_day * report_count_that_day / 2 ). Current solution gives reasonable 15min runtime, but it can probably be further improved.

PR #111, opened 2017-12-06
Add URL intelligence script

hellais:
This PR is about adding support for URL metadata including the category codes from the Citizenlab test-lists.

### DIFF
--- a/af/oometa/006-test-lists.install.sql
+++ b/af/oometa/006-test-lists.install.sql
@@ -2,9 +2,9 @@ BEGIN;
 
 select _v.register_patch( '006-test-lists', ARRAY[ '005-repeated-report' ], NULL );
 
-CREATE SEQUENCE IF NOT EXISTS cat_no_seq;
+CREATE SEQUENCE cat_no_seq;
 
-CREATE TABLE IF NOT EXISTS url_categories
+CREATE TABLE url_categories
 (
     cat_no INT NOT NULL default nextval('cat_no_seq') PRIMARY KEY,
     cat_code VARCHAR UNIQUE NOT NULL,
@@ -13,9 +13,9 @@ CREATE TABLE IF NOT EXISTS url_categories
     cat_old_codes VARCHAR
 );
 
-CREATE SEQUENCE IF NOT EXISTS url_no_seq;
+CREATE SEQUENCE url_no_seq;
 
-CREATE TABLE IF NOT EXISTS url
+CREATE TABLE url
 (
     url_no INT NOT NULL default nextval('url_no_seq') PRIMARY KEY,
     url VARCHAR,
@@ -28,9 +28,9 @@ CREATE TABLE IF NOT EXISTS url
     UNIQUE (url, country_no)
 );
 
-CREATE SEQUENCE IF NOT EXISTS country_no_seq;
+CREATE SEQUENCE country_no_seq;
 
-CREATE TABLE IF NOT EXISTS country
+CREATE TABLE country
 (
     country_no INT NOT NULL default nextval('country_no_seq') PRIMARY KEY,
     full_name VARCHAR UNIQUE NOT NULL,
--- a/af/oometa/006-test-lists.install.sql
+++ b/af/oometa/006-test-lists.install.sql
@@ -4,29 +4,32 @@ select _v.register_patch( '006-test-lists', ARRAY[ '005-repeated-report' ], NULL
 
 CREATE SEQUENCE cat_no_seq;
 
-CREATE TABLE url_categories
+CREATE TABLE url_category
 (
     cat_no INT NOT NULL default nextval('cat_no_seq') PRIMARY KEY,
     cat_code VARCHAR UNIQUE NOT NULL,
     cat_desc VARCHAR NOT NULL,
     cat_long_desc VARCHAR,
-    cat_old_codes VARCHAR
+    cat_old_codes VARCHAR[]
 );
+comment on table url_category is 'Contains the citizenlab URL category codes';
 
 CREATE SEQUENCE url_no_seq;
 
 CREATE TABLE url
 (
     url_no INT NOT NULL default nextval('url_no_seq') PRIMARY KEY,
-    url VARCHAR,
+    url VARCHAR NOT NULL,
     cat_no INT,
-    country_no INT,
-    date_added TIMESTAMP WITH TIME ZONE,
+    country_no INT NOT NULL,
+    date_added TIMESTAMP WITH TIME ZONE NOT NULL,
     source VARCHAR,
     notes VARCHAR,
-    active BOOLEAN,
+    active BOOLEAN NOT NULL,
     UNIQUE (url, country_no)
 );
+
+comment on table url is 'Contains information on URLs included in the citizenlab URL list';
 
 CREATE SEQUENCE country_no_seq;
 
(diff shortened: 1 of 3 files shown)


af/oometa/006-test-lists.install.sql
```diff
@@ -0,0 +1,42 @@
+BEGIN;
+
+select _v.register_patch( '006-test-lists', ARRAY[ '005-repeated-report' ], NULL );
+
+CREATE SEQUENCE IF NOT EXISTS cat_no_seq;
```
darkk: Drop `IF NOT EXISTS` please. The code should rather fail than continue in unexpected condition.

af/oometa/006-test-lists.install.sql
```diff
@@ -0,0 +1,42 @@
...
+
+CREATE TABLE IF NOT EXISTS url_categories
+(
+    cat_no INT NOT NULL default nextval('cat_no_seq') PRIMARY KEY,
```
darkk: FYI: the reason for me to create SEQUENCEs as separate DB objects was to reuse same SEQUENCE as a default value for different tables (e.g. temporary table and data table).

af/oometa/006-test-lists.install.sql
```diff
@@ -0,0 +1,42 @@
...
+    cat_code VARCHAR UNIQUE NOT NULL,
+    cat_desc VARCHAR NOT NULL,
+    cat_long_desc VARCHAR,
+    cat_old_codes VARCHAR
```
darkk: I'm not 100% sure, but you probably want it represented as array of strings.

af/oometa/006-test-lists.install.sql
```diff
@@ -0,0 +1,42 @@
...
+
+CREATE SEQUENCE IF NOT EXISTS country_no_seq;
+
+CREATE TABLE IF NOT EXISTS country
+(
+    country_no INT NOT NULL default nextval('country_no_seq') PRIMARY KEY,
+    full_name VARCHAR UNIQUE NOT NULL,
+    name VARCHAR UNIQUE NOT NULL,
+    alpha_2 VARCHAR(2) UNIQUE NOT NULL,
+    alpha_3 VARCHAR(3) UNIQUE NOT NULL
```
darkk: That's probably `char(2)` not `varchar(2)` and same for (3).
Or maybe it's DOMAIN like sha1.

af/oometa/006-test-lists.install.sql
```diff
@@ -0,0 +1,42 @@
...
+    UNIQUE (url, country_no)
+);
+
+CREATE SEQUENCE IF NOT EXISTS country_no_seq;
+
+CREATE TABLE IF NOT EXISTS country
+(
+    country_no INT NOT NULL default nextval('country_no_seq') PRIMARY KEY,
+    full_name VARCHAR UNIQUE NOT NULL,
+    name VARCHAR UNIQUE NOT NULL,
```
darkk: Do you need that indexes for `name` and `full_name`?... I'm trying to understand the reason behind them.

af/oometa/006-test-lists.install.sql
```diff
@@ -0,0 +1,42 @@
...
+CREATE TABLE IF NOT EXISTS url
+(
+    url_no INT NOT NULL default nextval('url_no_seq') PRIMARY KEY,
+    url VARCHAR,
+    cat_no INT,
+    country_no INT,
+    date_added TIMESTAMP WITH TIME ZONE,
+    source VARCHAR,
+    notes VARCHAR,
+    active BOOLEAN,
```
darkk: IMHO, lots of NOT NULLs are missing. E.g. for `url` and `cat_no`.

af/oometa/006-test-lists.install.sql
```diff
@@ -0,0 +1,42 @@
...
+
+CREATE TABLE IF NOT EXISTS country
+(
+    country_no INT NOT NULL default nextval('country_no_seq') PRIMARY KEY,
+    full_name VARCHAR UNIQUE NOT NULL,
+    name VARCHAR UNIQUE NOT NULL,
+    alpha_2 VARCHAR(2) UNIQUE NOT NULL,
+    alpha_3 VARCHAR(3) UNIQUE NOT NULL
+);
+
```
darkk: Please, add meaningful `comment on table foobar is 'foo is baring'`, these comments are really useful to visually distinguish forgotten temporary tables from "used production tables" in `\dt+`.

af/shovel/url_intelligence.py
```diff
@@ -0,0 +1,287 @@
...
+                      (name, full_name, alpha_2, alpha_3))
+
+def init_url_lists(working_dir, postgres, cat_code_no, country_alpha_2_no):
+    pgconn = psycopg2.connect(dsn=postgres)
+    with pgconn, pgconn.cursor() as c:
+        csv_glob = os.path.join(working_dir, 'test-lists', 'lists', '*.csv')
+        for csv_path in glob(csv_glob):
+            alpha_2 = os.path.basename(csv_path).split('.csv')[0].upper()
+            if alpha_2 == 'GLOBAL':
+                alpha_2 = 'XX' # We use XX to denote the global country code
```
darkk, line 120: Fun. `ZZ` is "Unknown county", `XX` is "Global". We should probably codify it somewhere.

af/oometa/006-test-lists.install.sql
hellais: Yes that makes sense. Will do that.

af/oometa/006-test-lists.install.sql
hellais: I change it to `CHAR` for the moment.

af/oometa/006-test-lists.install.sql
hellais: Ok

af/shovel/url_intelligence.py
hellais: `ZZ` being unknown is documented in OONI spec. `XX` is something internal and specific to the test-lists sync, but I guess we should still document it somewhere.
In any case it's in the ISO code namespace that is reserved for user assigned variables, so it should be safe.

af/oometa/006-test-lists.rollback.sql
```diff
@@ -0,0 +1,17 @@
...
+
+DROP TABLE url_category;
+
+DROP SEQUENCE IF EXISTS url_no_seq;
+
+DROP TABLE IF EXISTS url;
+
+DROP SEQUENCE IF EXISTS country_no_seq;
+
+DROP TABLE IF NOT EXISTS country;
```
darkk: Is there any reason to preserve this IFs?

hellais: It's slightly more convenient to have them for development and you want to do a partial rollback.

hellais: You should feel free to drop them when you deploy it.

af/shovel/url_intelligence.py
```diff
@@ -0,0 +1,303 @@
...
+
+def sync_repo(working_dir):
+    diffs = []
+    repo_dir = os.path.join(working_dir, 'test-lists')
+    if not os.path.isdir(repo_dir):
+        print("%s does not exist. Try running with --init" % repo_dir)
+        raise RuntimeError("%s does not exist" % repo_dir)
+    repo = git.Repo(repo_dir)
+    #previous_commit = repo.head.commit
+    #repo.remotes.origin.pull()
```
hellais: This should probably not be commented out.

hellais:
This is no longer relevant. We now have a citizenlab table which is populated using a similar script, but it's handled in a different part of the codebase.

We can close this and delete the branch.

PR #143, opened 2019-03-01
Delete all legacy pipeline code

hellais:
I propose that to reduce confusion as to what is the current pipeline code we delete all legacy code from the repo.

This makes this repo nice and tidy ✨


hellais:
@darkk did I miss something, do you think something else should be dropped (or did I drop something I shouldn't have?)

darkk:
I believe, there are other things that may be deleted as well.
- dead code from [`af/shovel/daily_workflow.py`](https://github.com/ooni/pipeline/blob/9a62ef13906032551f475af7d0fd70d2c32ab2e0/af/shovel/daily_workflow.py) (that was kept for `vimdiff` purposes during migration)
- /private directory (that is empty)
- s3://ooni-data/sanitised/* (that was kept to verify sanitisation code during migration)

hellais:
Thanks for following up!

> dead code from af/shovel/daily_workflow.py (that was kept for vimdiff purposes during migration)
> /private directory (that is empty)

I deleted these two as well in this branch

> s3://ooni-data/sanitised/* (that was kept to verify sanitisation code during migration)

I suggest we open a separate ticket for cleaning this up.

I filed: https://github.com/ooni/sysadmin/issues/283 for that

hellais:
@darkk if this looks good to you can you approve the PR and merge it?

darkk:
> I deleted these two

Sorry, I've failed to communicate the fact that _SOME_ parts of the file are dead while some [are not](https://github.com/ooni/pipeline/blob/1b2688d75a7abc09e446a7d965dd8011f5b5564d/af/shovel/autoclaving.py#L29). :disappointed:

So deletion of the whole file (instead of deletion of the dead code from the file) breaks `openobservatory/pipeline-shovel` build (as the file is referenced explicitly in `COPY` stage) and `autoclaving.py` script.

hellais:
Sounds good. Reverting that commit. @darkk let me know if it's good to merge now

darkk:
I see no obstacles to merging it.

bassosimone:
I've got a "go ahead" by both @darkk and @hellais

Issue #109, opened 2017-12-05
Generate daily dumps for ooni-explorer


hellais:
Currently in a fairly ad-hoc way I generated some static dumps for use by ooni-explorer.

We should ideally integrate this into the pipeline so that it is run on a daily basis when measurements are ingested.

In here the code snippet:

```python
import os
import psycopg2
import time
import json

def count_by_country(pgconn):
    results = []
    with pgconn, pgconn.cursor() as c: # main transaction
        c.execute('''SELECT COUNT(measurement.msm_no) as count,
    report.probe_cc AS probe_cc
    FROM measurement
    JOIN report ON report.report_no = measurement.report_no
    GROUP BY report.probe_cc;''')
        for count, probe_cc in c:
            results.append({
                'probe_cc': probe_cc,
                'count': count
            })

    return {
        'results': results
    }

def count_total(pgconn):
    total = -1
    with pgconn, pgconn.cursor() as c: # main transaction
        c.execute('SELECT COUNT(measurement.msm_no) as count FROM measurement')
        row = c.fetchone()
        total = row[0]

    return {
        'total': total
    }

def blockpage_count(pgconn):
    """
    We dump number of blocked sites / total tested per report that contain at
    least 1 blocked site for every country.
    We limit the query to 42, because that is how many charts fit in 3 pages on
    OONI Explorer.
    """
    query = '''
    SELECT block_count,
total_inputs,
test_start_time,
probe_cc FROM (
	SELECT
	SUM(CASE WHEN measurement.confirmed = true THEN 1 ELSE 0 END) as block_count,
	COUNT(input.input) as total_inputs,
	report.test_start_time as test_start_time,
	report.probe_cc AS probe_cc
	FROM measurement
	JOIN report ON report.report_no = measurement.report_no
	JOIN input ON input.input_no = measurement.input_no
	WHERE (report.test_name = 'web_connectivity' OR report.test_name = 'http_requests')
	GROUP BY report.test_start_time, report.probe_cc
)  as block_counts
WHERE block_count > 0
ORDER BY probe_cc, test_start_time ASC
'''
    results_by_country = {}
    with pgconn, pgconn.cursor() as c: # main transaction
        c.execute(query)
        for block_count, total_inputs, test_start_time, probe_cc in c:
            results_by_country[probe_cc] = results_by_country.get(probe_cc, [])
            results_by_country[probe_cc].append({
                'probe_cc': probe_cc,
                'block_count': block_count,
[...]

PR #144, opened 2019-03-14
OONI Pipeline workflow for API needed by OONI Explorer

hellais:
Here I will be pushing pipeline related code to create and update tables needed by the next gen OONI Explorer.


hellais:
The idea is that these tables will get updated by a daily job that runs after centrifugation.

In particular I am imagining we can incrementally update the stats tables with something like:

```
--- Daily workflow for ooexpl_bucket_msmt_count

--- Retrieve the last updated bucket and based on this value run the query below:
SELECT bucket_date
FROM ooexpl_bucket_msmt_count
ORDER BY bucket_date DESC
LIMIT 1;

INSERT INTO ooexpl_bucket_msmt_count (count, probe_asn, probe_cc, bucket_date)
SELECT 
COUNT(msm_no) as count,
probe_asn,
probe_cc,
bucket_date
FROM measurement 
JOIN report ON report.report_no = measurement.report_no
JOIN autoclaved ON autoclaved.autoclaved_no = report.autoclaved_no
WHERE bucket_date >= '2019-03-11'
GROUP BY bucket_date, probe_asn, probe_cc
ON CONFLICT (probe_asn, probe_cc, bucket_date) DO
UPDATE
SET count = EXCLUDED.count;

--- Daily workflow for ooexpl_recent_msmt_count

SELECT bucket_date
FROM ooexpl_recent_msmt_count
ORDER BY bucket_date DESC
LIMIT 1;

INSERT INTO ooexpl_recent_msmt_count (msmt_count, probe_cc, probe_asn, test_name, test_start_time, bucket_date)
SELECT
COUNT(msm_no) as msmt_count,
probe_cc,
probe_asn,
test_name,
test_start_time,
bucket_date
FROM measurement
JOIN report ON report.report_no = measurement.report_no
JOIN autoclaved ON autoclaved.autoclaved_no = report.autoclaved_no
WHERE 
bucket_date >= '2019-03-11'
AND test_start_time > current_date - interval '30 day'
GROUP BY probe_cc, probe_asn, test_name, test_start_time, bucket_date
ON CONFLICT (probe_cc, probe_asn, test_name, test_start_time, bucket_date) DO
UPDATE
SET msmt_count = EXCLUDED.msmt_count;;

-- END
```

hellais:
Relevant Pipeline ticket is: https://github.com/ooni/pipeline/issues/109

af/oometa/013-ooexplr-meta.install.sql
```diff
@@ -0,0 +1,81 @@
...
+probe_cc,
+probe_asn,
+test_name,
+test_start_time,
+bucket_date
+FROM measurement
+JOIN report ON report.report_no = measurement.report_no
+JOIN autoclaved ON autoclaved.autoclaved_no = report.autoclaved_no
+WHERE test_start_time > current_date - interval '31 day'
+GROUP BY probe_cc, probe_asn, test_name, test_start_time, bucket_date;
```
darkk: I'm unsure if it's practical to group by `test_start_time` as it's basically becomes "count measurements per report file". Is it really the granularity of the data needed for stats?
91% of there 593k rows in this table have `msmt_count = 1`.
E.g. grouping by `probe_cc, probe_asn, test_name, bucket_date` gives 86k rows (15% of original).
E.g. grouping by `probe_cc, probe_asn, test_name, bucket_date, date_trunc('hour', test_start_time)` gives 189k rows (32% of the original).
These two cases have less risk of "practically unbounded" growth.
The problem is that number of rows will further "degrade" as OONI Probe will be moving in "one measurement per report" direction, basically, ditching the goal of aggregation, if I understand it correctly.

af/oometa/013-ooexplr-meta.install.sql
```diff
@@ -0,0 +1,81 @@
...
+http_experiment_failure,
+dns_experiment_failure,
+control_failure,
+bucket_date
+FROM measurement
+JOIN input ON input.input_no = measurement.input_no 
+JOIN report ON report.report_no = measurement.report_no
+JOIN http_verdict ON http_verdict.msm_no = measurement.msm_no
+JOIN autoclaved ON autoclaved.autoclaved_no = report.autoclaved_no
+WHERE test_name = 'web_connectivity' AND measurement_start_time > current_date - interval '31 day';
```
darkk: It's a bit unclear, what's the long-term goal here. Is the goal to provide a 30d sliding window? Seems, it's not as "recent" is not the word that is used for the table name here. Is it a temporary ad-hoc workaround for ooni/api#59?

af/oometa/013-ooexplr-meta.install.sql
```diff
@@ -0,0 +1,81 @@
...
+CREATE TABLE ooexpl_bucket_msmt_count AS
+SELECT
+COUNT(msm_no) as count,
+probe_asn,
+probe_cc,
+bucket_date
+FROM measurement 
+JOIN report ON report.report_no = measurement.report_no
+JOIN autoclaved ON autoclaved.autoclaved_no = report.autoclaved_no
+GROUP BY bucket_date, probe_asn, probe_cc;
```
darkk: :bike: :derelict_house: &mdash; I'd suggest to use consistent naming patterns.
E.g. `msm` instead of `msmt` and `msm_count` instead of `msmt_count`.
E.g. `msm_count` instead of `count`.

af/oometa/013-ooexplr-meta.install.sql
hellais: The datums on this table will eventually be grouped on a per day basis. I think it's fine to also just consider measurement_start_time and not do this per report file aggregation, but it seemed like an easy optimisation at no cost of compromising the usage of data.

The problem with doing instead the aggregation on, say, a daily basis is that we then may have to recompute some stats or aggregation as new data from a bucket is processed.

darkk: > recompute some stats or aggregation as new data from a bucket is processed

I assumed that the goal of that intermediate representation and partial pre-aggregation is to speed-up reading by creating a `cube` of stats and making reading of whole table unnecessary when the only thing that is needed is the count of rows.

That's exactly "doing some more writes to speed-up reads".

That makes me think that optimizing for the data representation that is compact and reflects API performance needs is the goal. Counting rows in a reprocessed bucket after incremental reprocessing is a relatively cheap operation: it computes stats over the rows that are likely in the page cache and re-writes stats for the single bucket. Basically, that's poor-man partial update for materialized view having `bucket_date` as a shard key :)

af/oometa/013-ooexplr-meta.install.sql
hellais: Good point, I will look into fixing this.

af/oometa/013-ooexplr-meta.install.sql
hellais: Yes exactly. For the purpose of OONI Explorer we are going to in the initial MVP phase support showing the latest 30 days worth of measurements, but we may expand this to a larger window (though probably not larger than say 6 months), hence the name "recent".

If ooni/api#59 gets fixed it may make sense to drop this table, though maybe there is some performance tuning that we can do easier or better with a dedicated table?

af/oometa/013-ooexplr-meta.install.sql
hellais: Yeah you raise a good point here. It may be worthwhile as soon as we have consolidated all the stats we need to drop this table and consolidate it into one or more fact tables that are updated using the bucket_date key.

At least that is the principle or thought I had with creating, for example, the `ooexpl_bucket_msmt_count` table, which is as you say basically a poor-man partial update for mat view with bucket_date as shard key.

hellais: As suggested by @darkk we should probably group by truncated `measurement_start_time` using `date_trunc('day')`.

hellais:
TODO:
* [x] Replace msmt with msm
* [x] GROUP BY bucket_date and truncated measurement_start_time in recent_measurement table
* [x] Add comments to tables
* [x] Convert websites to a mat view
* [x] We should use the input_no instead of the input text on the ooexpl_websites table

To deploy it the steps are:
* [x] Add code in the centrifugation stage in the main transaction
* [x] To test it locally:
- Fetch a bucket of autoclaved data locally (it's ~5GB of storage) 
- Launch a postgresql instance via the script https://github.com/ooni/pipeline/blob/master/af/oometa/run-pg
- Then use similar snippets to debug it:
```
exec docker run $ttarg --interactive --rm \
    -v $PWD:/mnt:rw --workdir /mnt \
    $volargs \
    --privileged \
    --user $(id -u):$(id -g) \
    --memory 512M \
    --link spbmeta \
    openobservatory/pipeline-shovel:latest "$@"```
```

Memory limitation is there to resist the temptation of using too much memory early on.

```
for date in "$@"; do
    time ./centrifugation.py --start ${date}T00:00:00 --end $(date -d "$date + 1 day" +%Y-%m-%d)T00:00:00 --autoclaved-root /mnt/ac --postgres 'host=spbmeta user=oopguser'
done
```

hellais:
I ran the install script on hkgmetadb and it took 2279s to run, but then ran into a duplicate constraint violation.

As it turns out doing:
```
SELECT
COUNT(msm_no) as "count",
probe_cc,
probe_asn,
test_name,
date_trunc('day', measurement_start_time) as test_day,
bucket_date
FROM
[.. SNIP ..]
GROUP BY probe_cc, probe_asn, test_name, test_start_time, bucket_date, test_day;
```

Is not the same as:

```
GROUP BY 2,3,4,5,6;
```

I do wonder if this is a but in postgres or if this is expected behaviour and if so why?

af/oometa/013-ooexpl-tables.install.sql
```diff
@@ -0,0 +1,27 @@
...
+
+comment on table ooexpl_bucket_msm_count is 'OONI Explorer stats table for counting the total number of measurements since the beginning of time by probe_cc and probe_asn';
+
+CREATE TABLE ooexpl_daily_msm_count (
+    "count" bigint,
+    "probe_cc" character(2),
+    "probe_asn" integer,
+    "test_name" ootest,
+    "test_day" timestamp without time zone,
+    "bucket_date" date,
```
darkk, line 21: BTW, you could save some bytes here:

- `count` as a `bigint` (aka `int8`) does not make much sense, as `msm_no` is `integer` (aka `int4`). Yes, `msm_no` already takes 13% of available "address space", but 7x growth is still needed to take all the space, that'll take few years at a current rate.
-  using `date` for `test_day` saves another 4 byes.

Does it have measurable effect in terms of performance? It's unclear.

af/oometa/013-ooexpl-tables.install.sql
```diff
@@ -0,0 +1,27 @@
+BEGIN;
+
+select _v.register_patch( '013-ooexpl-tables', ARRAY[ '012-sha256-input-uniq' ], NULL );
+
+CREATE TABLE ooexpl_bucket_msm_count (
+    "count" bigint,
+    "probe_asn" integer,
+    "probe_cc" character(2),
+    "bucket_date" date,
```
darkk, line 9: Same story with `count` as an `int8` instead of `int4` applies here.

af/oometa/014-ooexpl-insert.rollback.sql
```diff
@@ -2,6 +2,6 @@ BEGIN;
 
 select _v.unregister_patch( '014-ooexpl-insert');
 
-TRUNCATE ooexpl_recent_msm_count, ooexpl_daily_msm_count;
+TRUNCATE ooexpl_bucket_msm_count, ooexpl_daily_msm_count;
```
darkk, line 5: BTW, should it be `TRUNCATE` or `DROP TABLE` ?

hellais: In this case TRUNCATE is correct, because this item is just about INSERTING into the table. The `DROP` instruction goes in the 013 `rollback` script.

darkk: Ah, true. Sorry, I was a bit confused.

hellais:
FTR these were the ooexpl-websites related tables:

`af/oometa/013-ooexpl-website-table.install.sql`

```
BEGIN;

select _v.register_patch( '013-ooexpl-website-table', ARRAY[ '012-sha256-input-uniq' ], NULL );

CREATE TABLE ooexpl_website_msm (
    msm_no integer PRIMARY KEY,
    input_no integer NOT NULL,
    probe_asn integer NOT NULL,
    probe_cc character(2) NOT NULL,
    measurement_start_time timestamp without time zone NOT NULL,
    anomaly boolean,
    confirmed boolean,
    failure boolean,
    bucket_date date
);

COMMIT;
```

`af/oometa/013-ooexpl-website-table.rollback.sql`
```
BEGIN;

select _v.unregister_patch( '013-ooexpl-website-table');

DROP TABLE ooexpl_website_msm;

COMMIT;
```

`014-ooexpl-website-insert.install.sql`
```
BEGIN;

select _v.register_patch( '014-ooexpl-website-insert', ARRAY[ '013-ooexpl-website-table' ], NULL );

INSERT INTO ooexpl_website_msm ("msm_no", "input_no", "probe_asn", "probe_cc", "measurement_start_time", "anomaly", "confirmed", "failure", "bucket_date")
SELECT
    measurement.msm_no,
    input.input_no,
    probe_asn,
    probe_cc,
    measurement_start_time,
    CASE
        WHEN blocking != 'false' AND blocking != NULL THEN TRUE
        WHEN measurement.msm_no IN (SELECT msm_no FROM http_request_fp) THEN TRUE
        ELSE FALSE
    END as anomaly,
    CASE
        WHEN measurement.msm_no IN (SELECT msm_no FROM http_request_fp) THEN TRUE
        ELSE FALSE
    END as confirmed,
    CASE
        WHEN control_failure != NULL OR blocking = NULL THEN TRUE
        ELSE FALSE
    END as failure,
    bucket_date
    FROM measurement
    JOIN input ON input.input_no = measurement.input_no
    JOIN report ON report.report_no = measurement.report_no
    JOIN http_verdict ON http_verdict.msm_no = measurement.msm_no
    JOIN autoclaved ON autoclaved.autoclaved_no = report.autoclaved_no
    WHERE test_name = 'web_connectivity'
    AND measurement_start_time > current_date - interval '31 day'
ON CONFLICT (msm_no) DO UPDATE SET
    anomaly = EXCLUDED.anomaly,
    confirmed = EXCLUDED.confirmed,
    failure = EXCLUDED.failure;

COMMIT;
```

`014-ooexpl-website-insert.rollback.sql`
```
BEGIN;

select _v.unregister_patch( '014-ooexpl-website-insert');

TRUNCATE ooexpl_website_msm;

COMMIT;
```

darkk:
FTR, I've updated `measurements` table in a following way:

```sql
metadb=# create unlogged table flags_fp_20190415 AS SELECT msm_no, true AS anomaly, true AS confirmed FROM http_request_fp;
SELECT 2758465

metadb=# create unlogged table flags_http_verdict_20190415 AS SELECT msm_no, true AS anomaly, NULL::bool AS confirmed FROM http_verdict WHERE blocking != 'false' AND blocking IS NOT NULL;
SELECT 16581201

metadb=# create unlogged table flags_20190415 AS SELECT msm_no, bool_or(anomaly) as anomaly, bool_or(confirmed) as confirmed FROM (select * from flags_http_verdict_20190415 UNION ALL select * from flags_fp_20190415) t GROUP BY msm_no;
SELECT 17241876
Time: 53s

metadb=# create unique index on flags_20190415 (msm_no);
CREATE INDEX
Time: 22s

metadb=# cluster flags_20190415 using flags_20190415_msm_no_idx;
CLUSTER
Time: 42s

metadb=# CREATE TABLE measurement_20190415 (
metadb(#     msm_no integer NOT NULL,
metadb(#     report_no integer NOT NULL,
metadb(#     frame_off public.size4 NOT NULL,
metadb(#     frame_size public.size4 NOT NULL,
metadb(#     intra_off public.size4 NOT NULL,
metadb(#     intra_size public.size4 NOT NULL,
metadb(#     measurement_start_time timestamp without time zone,
metadb(#     test_runtime real,
metadb(#     orig_sha1 public.sha1 NOT NULL,
metadb(#     id uuid NOT NULL,
metadb(#     input_no integer,
metadb(#     exc integer[],
metadb(#     residual_no integer NOT NULL,
metadb(#     msm_failure boolean,
metadb(#     anomaly boolean,
metadb(#     confirmed boolean
metadb(# );
CREATE TABLE
Time: 232.224 ms

metadb=# insert into measurement_20190415 select msm_no, report_no, frame_off, frame_size, intra_off, intra_size, measurement_start_time, test_runtime, orig_sha1, id, input_no, exc, residual_no, msm_failure, f.anomaly, f.confirmed from measurement left join flags_20190415 f using (msm_no);
INSERT 0 235250361
Time: 54m
metadb=# COMMENT ON TABLE measurement_20190415 IS 'Metadata for measurements stored within reports';
COMMENT
Time: 251 ms

metadb=# ALTER TABLE ONLY public.measurement_20190415 ADD CONSTRAINT measurement_v2_pkey PRIMARY KEY (msm_no);
CREATE INDEX measurement_v2_id_idx ON public.measurement_20190415 USING btree (id);
CREATE INDEX measurement_v2_input_no_idx ON public.measurement_20190415 USING btree (input_no);
CREATE INDEX measurement_v2_measurement_start_time_idx ON public.measurement_20190415 USING btree (measurement_start_time);
CREATE INDEX measurement_v2_report_no_idx ON public.measurement_20190415 USING btree (report_no);
ALTER TABLE
Time: 19m
CREATE INDEX
Time: 19m
CREATE INDEX
Time: 17m
CREATE INDEX
Time: 15m
CREATE INDEX
Time: 16m

-- these indexes were not used by the planner at both HKG and AMS instances anyway
metadb=# drop index measurement_anomaly_idx, measurement_confirmed_idx, measurement_exc_idx, measurement_msm_failure_idx, measurement_residual_no_idx;
DROP INDEX
Time: 420.535 ms
metadb=# begin; ALTER TABLE measurement RENAME TO measurement_20190415_old ; ALTER TABLE measurement_20190415 RENAME TO measurement; commit ;

```

hellais:
I am going to merge this so that future work can be based on top of this.

PR #145, opened 2019-03-18
Centrifugation docs

hellais:
As part of https://github.com/ooni/pipeline/pull/144 I wrote some docstrings to some of the stages of the centrifugation workflow.

I would like to have similar docstrings for other stages as well.

### DIFF
--- a/af/shovel/centrifugation.py
+++ b/af/shovel/centrifugation.py
@@ -177,7 +177,8 @@ def load_autoclaved_db_todo(pgconn, files, bucket):
     This functions tries to understand what needs to be done for a particular
     set of autoclaved filenames.
 
-    It looks at the code_ver stored in the autoclaved table and understands what action needs to be performed on a autoclaved file.
+    It looks at the code_ver stored in the autoclaved table and understands
+    what action needs to be performed on a autoclaved file.
 
     The possible actions are:
         * ingest ~ The autoclaved file does not appear in the DB, so it needs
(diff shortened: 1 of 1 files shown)


darkk:
I'd suggest to wrap docstrings at 120 cols as that's what github has as a default width for the review "window".

hellais:
> I'd suggest to wrap docstrings at 120 cols as that's what github has as a default width for the review "window".

Good call. I forgot to wrap a couple of docstrings. Fixed.

af/shovel/centrifugation.py
```diff
@@ -166,10 +173,35 @@ def load_autoclaved_index(autoclaved_index):
...
+        to be ingested for the first time. This creates new measurement
+        records.
+
+        * reingest  ~ The file hash in the autoclaved DB table has changed, so
+        it needs to be re-ingested. We preserve the `msm_no`, but update the
+        binary representation of the autoclaved file & re-parse data.
+
+        * reprocess ~ the code_ver in the autocalved table is older than that
+        of the centrifugation file, hence we need to re-parse the data (or some
+        subset of the tables)
```
darkk, line 194: I'm unsure what is the best option here while there are several:
- keep the description dense and add a link to [the corresponding paragraph of the doc](https://github.com/ooni/pipeline/blob/master/docs/pipeline-16.10.md#centrifugation)
- keep it as you wrote and reference the doc
- keep it as it currently is

I'm linking the doc to show that those terms are already defined somewhere else and it's unclear if the duplicate definition is needed.

Also, I'm unsure if _"update the binary representation of the autoclaved file"_ is actually correct and non-confusing way to describe the process that is going on. Yes, I agree, it's the phrase that was written before. Probably, the thing I meant was closer to _"update the database representation of the autoclaved file that has its binary representation changed"_ or to _"update the database representation of the autoclaved file"_. The thing I can say with confidence is that I've lost some sense while writing the text down. :-(

darkk (approved):
Thanks! That's good.

af/shovel/centrifugation.py
hellais: Ah it wasn't clear to me that https://github.com/ooni/pipeline/blob/master/docs/pipeline-16.10.md#centrifugation was in sync with the latest pipeline development (the last commit from you on it was in December 2017).

I would suggest we keep docs in the code (as it's always useful to have them close to the thing you are reading) and consolidate our pipeline documentation into a canonical set of documents (16.10.md + Readme.md, etc.) with an index.

darkk: Well, there were no significant design changes, so no updates to the "design doc" happened :-)

PR #149, opened 2019-04-05
Add small notes in headers of shovel files

hellais:
Sat down with @darkk at IFF and took some notes on some scripts.

Still need to clean up the more detailed notes and add those as well.


darkk (approved):
kewl!

PR #151, opened 2019-04-18
Make the confirmed flag more accurate

hellais:
Limit marking measurement with the confirmed flag only when the country
of the blockpage fingerprint is consistent with the measurements

### DIFF
--- a/af/shovel/centrifugation.py
+++ b/af/shovel/centrifugation.py
@@ -775,9 +775,9 @@ def calc_measurement_flags(pgconn, flags_tbl, msm_tbl):
                 true AS anomaly,
                 true AS confirmed
             FROM http_request_fp
-            JOIN fingerprint ON fingerprint.fingerprint_no = http_request_fp.fingerprint_no
-            JOIN measurement_meta ON http_request_fp.msm_no = measurement_meta.msm_no
-            JOIN report_blob ON report_blob.report_no = measurement_meta.report_no
+            JOIN fingerprint USING (fingerprint_no)
+            JOIN measurement_meta USING (msm_no)
+            JOIN report_blob USING (report_no)
             WHERE origin_cc = probe_cc
             AND http_request_fp.msm_no IN (SELECT msm_no FROM {msm})
             AND http_request_fp.msm_no >= %s AND http_request_fp.msm_no <= %s
--- a/af/shovel/centrifugation.py
+++ b/af/shovel/centrifugation.py
@@ -1542,7 +1542,7 @@ class HttpRequestFPFeeder(HttpRequestFeeder):
                     ''
                 ) from fingerprint
             ''')
-            assert '916446978a4a86741d0236d19ce7157e' == next(c)[0], 'fingerprint table does not match CODE_VER={}'.format(CODE_VER)
+            assert '63b78da421413ef8ae0d2f1555963e26' == next(c)[0], 'fingerprint table does not match CODE_VER={}'.format(CODE_VER)
             c.execute('SELECT fingerprint_no, body_substr, header, header_prefix, header_value FROM fingerprint')
             for fingerprint_no, body_substr, header, header_prefix, header_value in c:
                 if body_substr is not None:
--- a/af/shovel/centrifugation.py
+++ b/af/shovel/centrifugation.py
@@ -779,8 +779,8 @@ def calc_measurement_flags(pgconn, flags_tbl, msm_tbl):
             JOIN measurement_meta USING (msm_no)
             JOIN report_blob USING (report_no)
             WHERE origin_cc = probe_cc
-            AND http_request_fp.msm_no IN (SELECT msm_no FROM {msm})
-            AND http_request_fp.msm_no >= %s AND http_request_fp.msm_no <= %s
+            AND msm_no IN (SELECT msm_no FROM {msm})
+            AND msm_no >= %s AND msm_no <= %s
             UNION ALL
             SELECT msm_no, true AS anomaly, NULL AS confirmed FROM http_verdict
             WHERE msm_no IN (SELECT msm_no FROM {msm}) AND msm_no >= %s AND msm_no <= %s
--- a/af/shovel/centrifugation.py
+++ b/af/shovel/centrifugation.py
@@ -771,7 +771,7 @@ def calc_measurement_flags(pgconn, flags_tbl, msm_tbl):
         ''')
         c.execute('''
         INSERT INTO {flags} SELECT msm_no, bool_or(anomaly), bool_or(confirmed) FROM (
-            SELECT msm_no,
+            SELECT http_request_fp.msm_no AS msm_no,
                 true AS anomaly,
                 true AS confirmed
             FROM http_request_fp
@@ -779,8 +779,8 @@ def calc_measurement_flags(pgconn, flags_tbl, msm_tbl):
             JOIN measurement_meta ON http_request_fp.msm_no = measurement_meta.msm_no
             JOIN report_blob ON report_blob.report_no = measurement_meta.report_no
             WHERE origin_cc = probe_cc
-            AND msm_no IN (SELECT msm_no FROM {msm})
-            AND msm_no >= %s AND msm_no <= %s
+            AND http_request_fp.msm_no IN (SELECT msm_no FROM {msm})
+            AND http_request_fp.msm_no >= %s AND http_request_fp.msm_no <= %s
             UNION ALL
             SELECT msm_no, true AS anomaly, NULL AS confirmed FROM http_verdict
             WHERE msm_no IN (SELECT msm_no FROM {msm}) AND msm_no >= %s AND msm_no <= %s
--- a/af/shovel/centrifugation.py
+++ b/af/shovel/centrifugation.py
@@ -776,8 +776,8 @@ def calc_measurement_flags(pgconn, flags_tbl, msm_tbl):
                 true AS confirmed
             FROM http_request_fp
             JOIN fingerprint ON fingerprint.fingerprint_no = http_request_fp.fingerprint_no
-            JOIN measurement_blob ON http_request_fp.msm_no = measurement_blob.msm_no
-            JOIN report_blob ON report_blob.report_no = measurement_blob.report_no
+            JOIN measurement_meta ON http_request_fp.msm_no = measurement_meta.msm_no
+            JOIN report_blob ON report_blob.report_no = measurement_meta.report_no
             WHERE origin_cc = probe_cc
             AND msm_no IN (SELECT msm_no FROM {msm})
             AND msm_no >= %s AND msm_no <= %s
--- a/af/shovel/centrifugation.py
+++ b/af/shovel/centrifugation.py
@@ -771,8 +771,16 @@ def calc_measurement_flags(pgconn, flags_tbl, msm_tbl):
         ''')
         c.execute('''
         INSERT INTO {flags} SELECT msm_no, bool_or(anomaly), bool_or(confirmed) FROM (
-            SELECT msm_no, true AS anomaly, true AS confirmed FROM http_request_fp
-            WHERE msm_no IN (SELECT msm_no FROM {msm}) AND msm_no >= %s AND msm_no <= %s
+            SELECT msm_no,
+                true AS anomaly,
+                true AS confirmed
+            FROM http_request_fp
+            JOIN fingerprint ON (fingerprint_no)
+            JOIN measurement_blob USING (msm_no)
+            JOIN report_blob ON (report_no)
+            WHERE origin_cc = probe_cc
+            AND msm_no IN (SELECT msm_no FROM {msm})
+            AND msm_no >= %s AND msm_no <= %s
             UNION ALL
             SELECT msm_no, true AS anomaly, NULL AS confirmed FROM http_verdict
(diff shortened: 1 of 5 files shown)


hellais:
I am going to run:
```
CREATE UNLOGGED TABLE fingerprint_origin_20190418 AS
SELECT origin_cc, probe_cc, measurement.msm_no
FROM http_request_fp
JOIN fingerprint ON fingerprint.fingerprint_no = http_request_fp.fingerprint_no
JOIN measurement ON http_request_fp.msm_no = measurement.msm_no
JOIN report ON report.report_no = measurement.report_no;

CREATE UNLOGGED TABLE unconfirmed_20190418 AS
SELECT msm_no
FROM fingerprint_origin_20190418 
JOIN measurement USING(msm_no)
JOIN report ON report.report_no = measurement.report_no
WHERE fingerprint_origin_20190418.origin_cc != fingerprint_origin_20190418.probe_cc;

UPDATE measurement SET confirmed = NULL WHERE msm_no IN (SELECT msm_no FROM unconfirmed_20190418);
```

And once the CI converges merge and deploy this branch.

af/shovel/centrifugation.py
```diff
@@ -771,8 +771,16 @@ def calc_measurement_flags(pgconn, flags_tbl, msm_tbl):
...
+            SELECT http_request_fp.msm_no AS msm_no,
+                true AS anomaly,
+                true AS confirmed
+            FROM http_request_fp
+            JOIN fingerprint USING (fingerprint_no)
+            JOIN measurement_meta USING (msm_no)
+            JOIN report_blob USING (report_no)
+            WHERE origin_cc = probe_cc
+            AND http_request_fp.msm_no IN (SELECT msm_no FROM {msm})
+            AND http_request_fp.msm_no >= %s AND http_request_fp.msm_no <= %s
```
darkk: If I understand SQL correcly, you can replace `http_request_fp.msm_no` with bare `msm_no` in this case.
I'm unsure if it improves readability or worsens it. :)

af/shovel/centrifugation.py
```diff
@@ -771,8 +771,16 @@ def calc_measurement_flags(pgconn, flags_tbl, msm_tbl):
...
         INSERT INTO {flags} SELECT msm_no, bool_or(anomaly), bool_or(confirmed) FROM (
-            SELECT msm_no, true AS anomaly, true AS confirmed FROM http_request_fp
-            WHERE msm_no IN (SELECT msm_no FROM {msm}) AND msm_no >= %s AND msm_no <= %s
+            SELECT http_request_fp.msm_no AS msm_no,
+                true AS anomaly,
+                true AS confirmed
+            FROM http_request_fp
+            JOIN fingerprint USING (fingerprint_no)
+            JOIN measurement_meta USING (msm_no)
+            JOIN report_blob USING (report_no)
```
darkk, line 780: I'm unsure if it plays nicely with both invocations of `calc_measurement_flags()`. It probably does work for _ingestion_ case, but I'm unsure about reprocessing/reingestion case.

hellais: Can you describe a bit more what happens in the case of reprocessing/reingestion?

Would any of these tables not exist in those cases?

darkk: They will be probably empty WRT msm_no from `{msm}` table (if my memory serves me well). It's unclear if it'll lead to flag being cleared or just ignored. I'd suggest to test that use-case as well if you consider that use-case useful.

darkk:
@hellais are there cases when some measurement triggers fingerprints from different `origin_cc` ? Alike cases will confuse your query.

hellais:
> are there cases when some measurement triggers fingerprints from different origin_cc ? Alike cases will confuse your query

That is precisely the goal of the addition. For measurements in which we have wrongly stamped the `probe_cc` we don't want to be confirming them as cases of censorship but rather skip them.

For example the geoip of some probe may appear to be in `US`, but they are seeing a blockpage from `IR`. We don't want these to end up inside of the `US` country page as confirmed cases of censorship as it could freak people out.

Does this make sense?

af/shovel/centrifugation.py
hellais: I see. I guess this problem would have presented itself also without my addition though, right?

darkk:
@hellais sorry, I missed a word and confused you :) I meant the following:

> are there cases when some measurement triggers fingerprints from _SEVERAL_ different `origin_cc` ?

Given that you're running the query anyway you may also want to double-check if the results actually match with the expectations :)

af/shovel/centrifugation.py
darkk: I don't think so (as JOINing with a possibly-empty table produces empty set of rows while avoiding that does not have that outcome).

hellais:
Ah I see what you mean. Yes that can in fact happen, for example these fingeprints are identical but occur in different countries:
```
18 | SA |   | Server | Protected by WireFilter |  
19 | AE |   | Server | Protected by WireFilter |  
```

Another case is:

```
10 | FR | xtpage = "page-blocage-terrorisme" |   |   |  
14 | GF | xtpage = "page-blocage-terrorisme" |   |   |  
```

darkk:
So, if I understand the query correctly, there is a chance that you've just _unconfirmed_ some `confirmed`. It's unclear to me to me if those `confirmed` deserve _reconfirmation_ and if it's actually needed or it's me being mistaken :)

hellais:
Yes you are right and have a good point. I am looking into fixing that.

hellais:
For the moment I have reverted the `confirmed` flag un-setting. I will look into fixing it properly tomorrow.

hellais:
Ok so I found the problem with the original query and found a way of fixing it.

It seems like there are some "legitimate" cases in which perhaps we should in fact be confirming them (for example the RU blockpage in UA), but I am not sure we should be making a special case for those.

For the time being I will air on the side of caution and unconfirm stuff based on the following queries:

```
CREATE UNLOGGED TABLE fingerprint_origin_20190423 AS
SELECT origin_cc, probe_cc, measurement.msm_no
FROM http_request_fp
JOIN fingerprint ON fingerprint.fingerprint_no = http_request_fp.fingerprint_no
JOIN measurement ON http_request_fp.msm_no = measurement.msm_no
JOIN report ON report.report_no = measurement.report_no;


CREATE UNLOGGED TABLE unconfirmed_20190423 AS
SELECT msm_no 
FROM
(
	SELECT 
	msm_no,
	ARRAY_AGG(origin_cc) as origin_cc_array
	FROM fingerprint_origin_20190423 
	GROUP BY msm_no
) as m
JOIN measurement USING(msm_no)
JOIN report USING(report_no)
WHERE NOT probe_cc = ANY(m.origin_cc_array);

UPDATE measurement SET confirmed = TRUE WHERE msm_no IN (SELECT msm_no FROM unconfirmed_20190423);

-- UPDATE 73056
-- 1376.8 s
```

af/shovel/centrifugation.py
hellais: Can you give me some tips on how I might go about testing if the it poses some problems with reprocessing? I am not too familiar with how to operate the pipeline for reprocessing/reingestion.

darkk:
> unconfirm stuff

It's not trivial for me to understand if incremental corrections like that actually provide the expected result.
I'd suggest to build `flags` table out of used features according to the logic codified as a part of _centrifugation_ step (when it's ready) and do a pass verifying that it matches.
IMHO, spending few CPU-hours is more robust than hoping that human can spot a mistake during review :)

af/shovel/centrifugation.py
darkk: Sure. Sorry, it took me a while to understand what sort of tip may be useful :) Let me state some facts first:
- the query uses tables `measurement_meta` and `report_blob`
- `report_blob` is filled by `ReportFeeder`
- instance of `ReportFeeder` is created at `prepare_destination()`
- `ReportFeeder.report_row()` is called at `copy_data_from_autoclaved()` only `if code_ver is None` where `code_ver` comes from `iter_autoclaved_datum()`
- `core_ver` may be `NOT NULL` for measurements coming from `autoclaved` / `report` / `measurement` where `code_ver` in the table is less than `CODE_VER` declared in the code
- `NOTNULL(code_ver)` also calls `MeasurementExceptionFeeder` instead of `MeasurementFeeder` in `feeder = msm if code_ver is None else msm_exc`
- `calc_measurement_flags` called from `MeasurementExceptionFeeder` will get set of `msm_no` on "old" measurements (and only them)

It's unclear to me if joining "old" `msm_no` with "new" table `report_blob` will produce any output. I expect zero tuples to come out of that JOIN as I expect sets of `msm_no` values in these table to be non-intersecting.
So zero tuples will get `true AS confirmed` in `{flags}` table.
So I _expect_ `MeasurementExceptionFeeder.close()` to `UPDATE measurement` and drop all `confirmed` flags back to `NULL`. Maybe I'm wrong about that.

So, I suggest to test _at least_ the case when new table is added and reprocessing code path is triggered. The way I use to trigger that code path is ingesting a bucket locally, doing something like `UPDATE autoclaved SET code_ver = 1`, running `centrifugation.py` again. I'd suggest to compare flags set after the first and the second run. They _should_ be equal (as the data is not changed), but my understanding of the code makes me think that they will not be.

hellais: Thanks for explaining this it's very useful!

I think there are some good bits of insight and knowledge that deserve going into the pipeline docs.

The last bit on updating the code_ver and re-running the centrifugation is especially useful and I am going to see if I can codify it as an integration test.

darkk: `tsvdump` / `tsvdiffstat` were some of the proto-test scripts those may be useful for the case.

hellais: According to the test the confirmed and unconfirmed count stays constant before and after reprocessing:
https://travis-ci.org/ooni/pipeline/builds/524104818#L787
https://travis-ci.org/ooni/pipeline/builds/524104818#L894

I still need to fully understand the implications of the text you are written above and wrap my head around it, but it seems like it passes the basic test.

Can you think of something more tricky that I can do to check more in depth if things are working as expected?

How can I test, for example, for re-ingestion?

darkk: > count stays constant

Indeed. I'm a bit surprised and still don't understand why it works though. Maybe that's a symptom of some other unexpected behavior in the code...

I'd also suggest to consider checking _anomaly_ counts as well. The check comes at a cost of few lines of code and almost no extra runtime and these two flags have slightly different semantics and, maybe, something will be eventually caught.

> something more tricky

IDK. You can disable `FLAG_TRUE_TEMP` and inspect tables manually. Or set a `pdb` breakpoint, do something like `COMMIT; BEGIN;` and inspect the tables in a concurrent connection.

> re-ingestion

The easiest way to trigger re-ingestion code path is to `UPDATE autoclaved SET file_sha1 = digest('wat', 'sha1')`.

hellais: I added a check for the anomaly flag value as well and it succeeds.

I also added a check that triggers re-ingestions and does the check there too and it also leads to consistent results.

hellais:
> I'd suggest to build flags table out of used features according to the logic codified as a part of centrifugation step (when it's ready) and do a pass verifying that it matches.

I started the following query and will validate it against the msm flags once it converges.

```
CREATE UNLOGGED TABLE flags_centrifugation_check_20190426 AS
SELECT msm_no, bool_or(anomaly) as anomaly, bool_or(confirmed) as confirmed FROM (
	SELECT http_request_fp.msm_no AS msm_no,
	    true AS anomaly,
	    true AS confirmed
	FROM http_request_fp
	JOIN fingerprint USING (fingerprint_no)
	JOIN measurement USING (msm_no)
	JOIN report USING (report_no)
	WHERE origin_cc = probe_cc
	UNION ALL
	SELECT msm_no, 
		true AS anomaly, 
		NULL AS confirmed 
	FROM http_verdict
	WHERE blocking != 'false' AND blocking IS NOT NULL
	) t GROUP BY msm_no;
```

hellais:
Some CPU cycles later:
```
CREATE UNLOGGED TABLE flags_msm_check_20190426 AS
SELECT msm_no, anomaly, confirmed
FROM measurement
WHERE anomaly = TRUE OR confirmed = TRUE;

SELECT COUNT(*) FROM flags_msm_check_20190426 WHERE confirmed = TRUE;
-- 2683965

SELECT COUNT(*) FROM flags_centrifugation_check_20190426 WHERE confirmed = TRUE;
-- 2683965

SELECT msm_no
    FROM flags_msm_check_20190426 a
    JOIN flags_centrifugation_check_20190426 b 
        USING (msm_no)
    WHERE a.confirmed = TRUE AND
    a.confirmed != b.confirmed;
-- No rows
```

The confirmed flags seem to be consistent. The `anomaly` flags are not, but I suspect this may have to do with some other work related to anomaly flags.
I believe at some point it would be wise retrigger reprocessing of all measurements and redo all these flags on all measurements.

In any case based on the above checks I think it's reasonable to deploy this.

@darkk thanks for the tips on how to test all of this.

darkk:
Just a reminder: `((true = true) and (true != null))` is `null` while `((true = true) and (true is distinct from null))` is `true`, so I'm unsure if you got expected results from the comparison.
 :man_shrugging:
I filled a ticket #162 so it's not forgotten.

PR #154, opened 2019-04-26
Add block page fingerprint for AF

anadahz:
Example measurement:
https://explorer.ooni.torproject.org/measurement/20171130T100203Z_AS38742_DmzL8KtUBadHNFU6wchlAgOy7MBwpDH75ZWdNzs0e55ArgFffx?input=http:%2F%2Fhowtogrowmarijuana.com


hellais:
I moved this into an issue, because you actually need to create a dedicated migration script and reprocess the measurements in order for it to work.

We should probably be doing it in batch at some point once @darkk finishes writing docs for it.

anadahz:
Ack, I can set the fingerprints as issues.

Issue #137, opened 2019-01-23
Define metrics and alarms regarding "failed" measurements


hellais:
This entails:

* [ ] Defining how this metric looks like
* [ ] Adding to the pipeline documentation about this failed metric
* [ ] Defining a procedure for how these failed measurements should be handled

darkk:
As it was [recently highlighted](https://github.com/ooni/pipeline/pull/177#discussion_r279317109) the pipeline works in _garbage-in-garbage-out_ mode trying to give a _best-effort_ service to incoming measurements.

That effectively means, that pipeline tries hard to ingest everything barely ingestable and takes a note if something went wrong. The places where the notes are taken are (at least) the following:
- `measurement.exc` column storing array of hashes of python exceptions happened during ingestion
- `measurement.residual_no` column storing a reference to `residual` table recording a weak "schema" of "unknown" json fields
- `measurement.msm_failure` that is currently unused, the idea was to mark measurement like #153 with this flag
- `badblob` table storing `badblob` records from `autoclaved/YYYY-MM-DD/index.json.gz`, that represent exceptions happening during _autoclaving_ stage
- `badmeta` table should store exceptions happening while ingesting very basic measurement metadata to `autoclaved`, `report` and `measurement` tables
- `badrow` stores record that PostgreSQL refuses to accept as a `COPY` input
- Airflow TaskInstance / DAG state, [e.g. there are some buckets from 2016](http://127.0.0.1:8080/admin/taskinstance/?flt0_dag_id_equals=hist_canning&flt1_state_equals=failed&flt4_task_id_equals=meta_pg) those are not ingested into PostgreSQL at all

_Ideally_ all of those records should be clean, that's the target for the metric, but code is imperfect. `badmeta` has no samples so far and it was the reason to introduce `FLAG_DEBUG_CHAOS` to debug that code path. Also, `badrow` table may deserve a patch that will record `bucket_date` as a part of a "bad row" to ease triaging (but it's rare error, so it did not happen so far). Also, Airflow sometimes fails to deliver expected logs via web-interface, so `find /srv -ipath '*/meta_pg/2016-02-23T00:00:00'` sometimes is the way to read logs at `datacollector.ingra.ooni.io`.

Here are few more lines on handling `measurement.exc` as an example.

[...]

PR #177, opened 2019-04-29
Add extractor for the telegram test

hellais:
@darkk when the docs are ready I would like to try rolling this out.

### DIFF
--- a/af/shovel/centrifugation.py
+++ b/af/shovel/centrifugation.py
@@ -1708,7 +1708,7 @@ class TelegramFeeder(BaseFeeder):
 
         http_blocking = t.get('telegram_http_blocking', None)
         tcp_blocking = t.get('telegram_tcp_blocking', None)
-        return '{:d}\t{:d}\t{}\t{:d}\t{}\t{}\t{}\t{}\t{}\n'.format(
+        return '{:d}\t{}\t{}\t{}\t{}\n'.format(
                     msm_no,
                     pg_quote(web_failure),
                     pg_quote(web_blocking),
--- a/af/shovel/centrifugation.py
+++ b/af/shovel/centrifugation.py
@@ -1028,11 +1028,11 @@ def calc_measurement_flags(pgconn, flags_tbl, msm_tbl):
             FROM http_request_fp
             JOIN fingerprint USING (fingerprint_no)
             WHERE msm_no IN (SELECT msm_no FROM {msm})
-            AND msm_no >= %s AND msm_no <= %s
+            AND msm_no >= {msm_min} AND msm_no <= {msm_max}
 
             UNION ALL
             SELECT msm_no, true AS anomaly, NULL AS confirmed FROM http_verdict
-            WHERE msm_no IN (SELECT msm_no FROM {msm}) AND msm_no >= %s AND msm_no <= %s
+            WHERE msm_no IN (SELECT msm_no FROM {msm}) AND msm_no >= {msm_min} AND msm_no <= {msm_max}
               AND blocking != 'false' AND blocking IS NOT NULL
 
             UNION ALL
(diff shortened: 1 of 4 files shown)


af/shovel/centrifugation.py
```diff
@@ -1273,6 +1273,46 @@ def pop(datum):
         for key in known_keys:
             t.pop(key)
 
+class TelegramFeeder(BaseFeeder):
+    min_compat_code_ver = 4
```
hellais: I think I need to bump this up by 1.

hellais: @darkk Should this be set to 5 or 6?

hellais: I set it to 6.

af/shovel/centrifugation.py
```diff
@@ -1273,6 +1273,46 @@ def pop(datum):
...
+        web_status = t.get('telegram_web_status', None)
+        web_blocking = None
+        if web_status == 'ok':
+            web_blocking = False
+        elif web_status == 'blocked':
+            web_blocking = True
+
+        http_blocking = t.get('telegram_http_blocking', None)
+        tcp_blocking = t.get('telegram_tcp_blocking', None)
+        return '{:d}\t{}\t{}\t{}\t{}\n'.format(
```
hellais: I had made a typo here in the number of columns to write into the DB, yet no exception was thrown.

@darkk is there reason to swallow these kinds of exceptions? Should maybe an issue be opened to consider at least logging them?

hellais: From slack:
```
Yes, garbage-in-garbage-out.
That's done that way so the malformed measurement is still indexed and marked with `exc` and does not block the pipeline.
You may enable `FLAG_FAIL_FAST` to avoid exceptions suppression.
```

hellais: More context from slack:
```
Storing exception as 32bit hashsum may sound like a strange idea, that's done to prevent database bloat in a pathological case as 1M of traces is not something trivial to deal with.
That idea is to use the hashsum to triage "failures" and pay attention to "popular" failures first.
```

darkk: FTR, the hash is computed by [`exc_hash()`](https://github.com/ooni/pipeline/blob/065cccdfeb531e93a22d2aacc05ec05e990f99ee/af/shovel/centrifugation.py#L996-L1006).

af/shovel/centrifugation.py
```diff
@@ -1273,6 +1273,46 @@ def pop(datum):
...
+
+    @staticmethod
+    def pop(datum):
+        if datum['test_name'] != 'telegram':
+            return
+        t = datum['test_keys']
+        known_keys = ('telegram_web_failure', 'telegram_http_blocking',
+                      'telegram_web_status', 'telegram_tcp_blocking')
+        for key in known_keys:
+            t.pop(key)
```
darkk, line 1752: `web_status` values besides `ok` and `blocked` are not parsed, but the key is pop'ed despite being potentially malformed. That may be a bad idea depending on circumstances.

hellais: What is the best practice for handling cases of malformed data? Should I be throwing an exception?

af/shovel/centrifugation.py
hellais: I documented this as part of an issue so we don't forget: https://github.com/ooni/pipeline/issues/185.

hellais:
The pipeline doc comment that is most relevant to rolling this out: https://github.com/ooni/pipeline/pull/183/files#r279376406.

af/shovel/centrifugation.py
darkk: IIRC, I tried to follow the following pattern: if the key has unknown (malformed?) value then the code does not `pop` it. So it ends in `residual` and marks the measurement as a measurement deserving some attention within a processes discussed in #137.

af/shovel/centrifugation.py
darkk: That's correct.

af/shovel/centrifugation.py
hellais: I see. And how do handle checking the validity of the value given that the conversion logic goes inside of the getter?

darkk: Other feeders follow the pattern when the every present field is validated at `row()` and an exception (or assertion) is raised if the field has an invalid value.

hellais:
We are moving this forward by integrating this directly into the fastpath: https://github.com/ooni/pipeline/pull/242

Issue #150, opened 2019-04-08
Document CODE_VER_REPROCESS in centrifugation.py


hellais:
In the centrifugation.py script there is a variable called CODE_VER_REPROCESS.

From chatting with @darkk we agreed that it should be better explained and if it's not in fact   CODE_VER_REINGESTION.

hellais:
@darkk can you clarify the difference between these two flags?

There is no mention of these flags in the reprocessing doc.

darkk:
@hellais can you clarify, what flags do you mean?
There is no such thing as `CODE_VER_REINGESTION`.
As far as I understand `CODE_VER_REPROCESS` is described accurately.

hellais:
I think this was something that came up when we were chatting about this in person. I recall that when I showed the line of code with `CODE_VER_REPROCESS` in it (see: https://github.com/ooni/pipeline/blob/master/af/shovel/centrifugation.py#L43) you were unsure of the purpose and were wondering if the meaning of this flag was in fact to trigger REPROCESSING of measurements or if i was meant to trigger REINGESTION.

Do you remember this discussion?

I guess the question here is if you could clearly state the purpose of this flag in the code and when and if somebody should be changing it or what is the meaning of having it in the code.

darkk:
Yep, I remember that discussion.
I re-read relevant code and, seems, it does not trigger reingestion and everything is named correctly.
The purpose is to have a constant defined as "the oldest possible `CODE_VER` value".
As far as I see, setting the `autoclaved.core_ver` to `0` has no special meaning (unless I'm mistaken).
Maybe, [`UPDATE autoclaved SET code_ver = 1`](https://github.com/ooni/pipeline/blob/master/tests/test_integration.py#L182) should rather be replaced with `code_ver = 0` to put those two magic values in sync. There should be no difference as `min(min_compat_code_ver)` is `2`.

hellais:
Excellent, thanks for clarifying that. I think we can consider this issue resolved then.

Issue #135, opened 2019-01-23
Document how to partially reprocess measurements


hellais:
There are times in which we may make some changes to the centrifugation stage and we may want to reprocess a subset of the measurements.

It's useful to document how to do this and update the `Reamde.md` to reflect.

These docs should also include reference to how a "fast path" for some measurements can be achieved.

darkk:
[Report deletion](https://github.com/ooni/pipeline/blob/master/docs/delete-report.md) is also relevant to this ticket and is documented in the linked note.

gabelula:
@darkk Are you working on this one? We need more documentation on this issue.

hellais:
This is documented inside of https://github.com/ooni/pipeline/pull/183 and those docs are really useful!

Thanks for putting them together.

Issue #134, opened 2019-01-23
Document re-ingestion of data in pipeline


hellais:
Checklist:

* [x] Verify that re-ingestion and reprocessing work as expected
* ~[ ] document the procedure at HOWTO level~
* [x] document unexpected side-effects (there are some)

darkk:
I believe that #183 mostly addresses the issue. Unfortunately it's not a HOWTO-level, but checklist-level due to unclear ways to predict pathological performance.

hellais:
@darkk there are some pending comments on that PR that I would like to have at least be addressed before this is considered closed.

hellais:
@darkk can you speak about the "unexpected side-effects (there are some)" of reprocessing the measurements?

darkk:
Some are described in [Preface section of the doc](https://github.com/ooni/pipeline/blob/dbd4ea463efc2137490e18812185b0a7849712b0/docs/reprocess-report.md#preface). I've mainly meant that blunt reprocessing of all the tables would produce lots of unnecessary updates to PG database, so, `code_ver` and `min_compat_code_ver` are introduced.
I've also meant that following _fast path_ and updating measurement flags using the features manually may lead to inconsistencies, so it should be done in-sync with code. You've also seen that during #151, that's why #162 appeared.
That's mostly it :-)

hellais:
As we the howto level tutorial was placed out of scope I think we can consider this issue solved.

Issue #133, opened 2019-01-23
Document writing feature extractors for new tests


hellais:
This is about documenting how (and when!) to add “feature extractors” for nettests that are currently not supported and how to reprocess the data to populate the metadb.

Related issue: https://github.com/ooni/api/issues/61

hellais:
I think that this PR https://github.com/ooni/pipeline/pull/177 plus what is documented inside of the reprocessing and reingestion doc (https://github.com/ooni/pipeline/pull/183) is enough for me write documentation on how to write new feature extractors for tests.

I am un-assigning this from @darkk and assigning it to myself so I can finish writing it as I rollout https://github.com/ooni/pipeline/pull/177.

HarendraSingh22:
Anyone working on this?

PR #183, opened 2019-04-29
Doc on reprocessing & reingestion

darkk:
I believe, the document answers some of the questions of #133, #134, #135, #150.


docs/reprocess-report.md
```diff
@@ -0,0 +1,122 @@
+This document describes how to reprocess historical data ingesting new features while minimizing resources waste and negative effect on the on-going data processing.
+
+This document is valid as of March 2019.
```
hellais: You may want to bump the date up to April 2019.

darkk: May 2019 also sounds valid :-)

docs/reprocess-report.md
```diff
@@ -0,0 +1,122 @@
...
+- update `fingerprint` table in [the database schema](https://github.com/ooni/pipeline/blob/065cccdfeb531e93a22d2aacc05ec05e990f99ee/af/oometa/) following [an example](https://github.com/ooni/pipeline/blob/065cccdfeb531e93a22d2aacc05ec05e990f99ee/af/oometa/003-fingerprints.install.sql#L31-L62) and [roll it out](https://github.com/ooni/sysadmin/blob/4defab8e92a2e53e2679a17214162ed058089e7f/ansible/deploy-pipeline-ddl.yml). The fingerprints are stored in the schema to generate `fingerprint_no serial`.
+- create a temporary table having `msm_no` of the measurements matching the fingerprint _with confidence_ according to the _derived_ features existing in database (e.g. `select msm_no from http_request where body_sha256 = '\x833b2fb8887eed1c0d496670148efa8b6a6e65b89f8df42dbd716464e3cf47a6'` for static blockpages)
+- insert those `msm_no` together with matching `fingerprint_no` into `http_request_fp` table as if those were actually ingested by `centrifugation.py`
+- update `anomaly` and `confirmed` flags in `measurement` table for the affected measurement according to the logic codified in `calc_measurement_flags()`
```
hellais, line 46: There is quite some amount of tricky manual work involved in doing this. Do you think there is a way of doing this in a more automated way, but that requires less human input (which may be error prone)?

docs/reprocess-report.md
```diff
@@ -0,0 +1,122 @@
...
+
+This case is almost the same one as the case of a static blockpage: the MetaDB has all the data to follow fast-path updating measurement metadata (`http_request_fp` table, `confirmed` and `anomaly` flags, etc.) with direct DB queries. The downside of fast-path is that it'll lead to duplication of logic between the queries and `centrifugation.py` that may (by mistake) lead to inconsistencies if the logic is not perfectly equivalent.
+
+Overall steps needed to mark existing & future measurements are the same as for HTML blockpage fingerprint with small alterations:
+
+- ...
+- create a temporary table having `msm_no` of the measurements matching the fingerprint _perfectly_ according to features existing in database (e.g. `select msm_no from http_request where headers->>'Location' = 'http://homeline.kg/access/blockpage.html'`, keep in mind that keys of headers are case-sensitive)
+- ...
```
hellais: I guess here there is a missing step which is that of actually updating the `anomaly` and `confirmed` column flags with the result of the above `msm_no`?
I think it's useful to provide a concrete example even if it will become not 100% accurate as the code changes to give an idea of how this can be achieved.

docs/reprocess-report.md
```diff
@@ -0,0 +1,122 @@
...
+- update `centrifugation.py`: set up-to-date `fingerprint` table checksum in `HttpRequestFPFeeder.__init__()`. There is no need to bump `CODE_VER` for feature-based fingerprints as we are 100% confident that reprocessing is not needed and ongoing data processing is paused.
+- ...
```
hellais: What is this extra step here?

docs/reprocess-report.md
```diff
@@ -0,0 +1,122 @@
...
+- Bump `CODE_VER`, set `min_compat_code_ver` for the new feeder
+- `TheFeeder.row()` creates a string that is suitable for sending to the table via `COPY`
+- `TheFeeder.pop()` removes fields from the JSON object those are completely ingested by the feeder and should NOT be considered a part of the _residual_
+- test, deploy, reprocess all (or the affected) buckets under GNU Make control
+
+One may save significant amount of CPU time marking old _autoclaved_ files as already processed by the new version of code bumping their corresponding `code_ver` in the database. It may be useful in a case when a feature has to be extracted **only(!)** from a known subset of reports, so the reports that have no data on the specific feature may be skipped safely. Example is extracting a feature of a "low-volume" test. E.g. `web_connectivity` test takes 99.4% of data volume of 2019Q1, so _any_ other test is a low-volume one. Another example is a extracting a feature that was shipped as a part of some specific `software` version, so `autoclaved` having no records coming from the new software may be manually labeled with a newer `code_ver` and skipped safely.
```
hellais, line 90: Can you provide a concrete code example of how to do this?

For example suppose I want to apply this CPU saving to deploy/rollout https://github.com/ooni/pipeline/pull/177

docs/reprocess-report.md
```diff
@@ -0,0 +1,122 @@
...
+Reingestion is different from reprocessing as it may handle changes to autoclaved files themselves and update `autoclaved`, `report`, `measurement` tables accordingly. The easiest way to force reingestion manually is to set `autoclaved.file_sha1` to all-zeros of something like `digest('', 'sha1')`. One of the possible reasons for that is [report deletion](./delete-report.md).
+
+## GNU Make crutch for Airflow
+
+Airflow has an issue in a scheduler, it starts consuming unreasonable amount of resources if there are thousands of _running_ DAGs. So, reprocessing of ≈2300 daily buckets of OONI data has to be micro-managed. One of the usual Linux tools to execute parallel processes is GNU Make, so, it was taken for the [`pipeline-reprocess`](https://github.com/ooni/sysadmin/blob/4defab8e92a2e53e2679a17214162ed058089e7f/scripts/pipeline-reprocess) script.
+
+The way to use the script is the following:
+
+- download it to your $HOME at `datacollector.infra.ooni.io` running Airflow
+- edit `PRJ` with a slug representing a reprocessing session
```
hellais: Can you provide an example of what is meant by this?

docs/reprocess-report.md
```diff
@@ -0,0 +1,122 @@
...
+- choose a way to list buckets-to-reprocess with `TYPEOF_DEPS`
```
hellais, line 130: Can you provide an example of what is meant by this?
I see the docstring says `# sed-deps, af-deps, or metadb-deps -- something like that`, but I don't understand the different between these or why I should be using one or the other.

docs/reprocess-report.md
```diff
@@ -0,0 +1,122 @@
...
+- edit `$(PRJ)/...-deps` to reflect the desired logic to select buckets to reprocess
```
hellais: Is this a necessary step in most cases? Is it wise to have this logic be editing on the fly on the machine and not checked into version control?

docs/reprocess-report.md
darkk: Here is an example: https://github.com/ooni/sysadmin/commit/bf9c967da7b2e2cc0c5efca0351cdf679d861b2f#diff-0bd5c245bef1335609715572487a3117

docs/reprocess-report.md
darkk: Here is the reference: https://github.com/ooni/sysadmin/blob/master/scripts/pipeline-reprocess#L55-L84
`sed-deps` refers to generation of $(PRJ)/sed-deps from affected.bucket_date by means of `sed`. `metadb-deps` is alike example using MetaDB as a data source, `af-deps` uses Airflow database as a data source for the list of buckets to reprocess.

docs/reprocess-report.md
darkk: IMHO, it has too many moving parts WRT performance and scope of systems touched to be "easily" automated.
I think, several manual cycles through the checklist with some variety of inputs are needed before trying to automate it.

docs/reprocess-report.md
darkk: I decided that `...` is more readable than a verbatim copy-paste from the previous paragraphs, so I've written down _small alterations_ to attract the attention to them.

docs/reprocess-report.md
darkk: As it's said in the paragraph before the list:

> Overall steps needed to mark existing & future measurements are the same as for HTML blockpage fingerprint

So it's  _"roll out `openobservatory/pipeline-shovel` and unpause data ingestion"_.

Can you suggest a better way to represent the same idea without distracting the reader with repetition?

docs/reprocess-report.md
darkk: #177 touches only `test_name = 'telegram'` and we don't need to reprocess the files those are already current, so it's something close to
```sql
UPDATE autoclaved SET code_ver = 6 WHERE code_ver = 5 AND autoclaved_no IN (
    SELECT autoclaved_no FROM autoclaved WHERE autoclaved_no NOT IN (
        SELECT DISTINCT autoclaved_no FROM report WHERE test_name = 'telegram'));
```

docs/reprocess-report.md
darkk: Yes, it's necessary to keep every `*.tiflag` file in a separate directory. All in all, GNU Make operates on files.

Indeed, there is some value in checking in the edited file into version control to preserve the history, but there is no other value. So, I assume that _committing for historical purposes_ is an assumed action. Otherwise we should write "and commit to version control, and push, and eject the device safely before unplugging, and use `shutdown -h` instead of pulling the cord". All those actions are good and, probably, needed, but those actions are not highly relevant to the discussed matter.

I say "download it to your $HOME at `datacollector.infra.ooni.io` running Airflow" to name the machine and say that it does not need root (and calls `sudo`).

docs/reprocess-report.md
hellais: If that is what you mean then I suggest you repeat the sentence. It's better to have some repetition, but not be ambiguous.

docs/reprocess-report.md
hellais: I understand now what you meant, I had missed that gotcha in the first sentence here.
I suggest repeating the steps or at the very least stating which steps from the "new HTML blockpage fingerprint" need to repeated in here.

docs/reprocess-report.md
hellais: I see. Is there some recommendations for how to go about adding fingerprints to the DB in batch?

We currently have 46 issues open with the label fingerprintdb (https://github.com/ooni/pipeline/issues?q=is%3Aopen+is%3Aissue+label%3Afingerprintdb) and going through all this dance to ingest each of them is going to require a non-trivial amount of time.

I am happy to tradeoff some performance (CPU time or IO time) for less human time.

darkk: The only recommendation I can give is the same: just treat the checklist as a declarative, not an imperative text, so, perform a batch of actions for every step. You're right being worried, naive non-batch implementation of the checklist for every ticket will take ~11.5 years to converge.

gabelula:
@hellais Do you think we can merge this pull request and close related issues?

hellais:
> @hellais Do you think we can merge this pull request and close related issues?

I still need to do another pass of reviewing it based on the latest comments from @darkk. As soon as I have done so I will merge it and close the issues.

docs/reprocess-report.md
hellais: Thanks. I added this as a concrete example in the text.

hellais:
I have done another pass at reviewing the text, integrated some of the comments from @darkk inside of it and to me it's good to merge.

Thank you so much for putting it together!

As soon the CI converged I shall merge it.

PR #198, opened 2019-05-31
Ooexpl wc confirmed

hellais:
Create table "ooexpl_wc_confirmed" with columns: confirmed_count, msm_count, test_day, probe_cc, probe_asn.

confirmed_count is the aggregated count of confirmed measurements over 2 months

### DIFF
--- a/af/shovel/centrifugation.py
+++ b/af/shovel/centrifugation.py
@@ -1685,6 +1685,34 @@ ON CONFLICT (probe_cc, probe_asn, test_name, test_day, bucket_date) DO
 UPDATE
 SET count = EXCLUDED.count;''', [bucket])
 
+    # Materialized view query:
+    """
+CREATE MATERIALIZED VIEW ooexpl_wc_confirmed AS
+SELECT
+COALESCE(SUM(CASE WHEN confirmed = TRUE THEN 1 ELSE 0 END), 0),
+COUNT(*),
+date_trunc('day', test_start_time),
+probe_cc,
+probe_asn
+FROM (
+	SELECT
+	DISTINCT input as input,
+	test_start_time,
+	probe_cc,
+	probe_asn,
+	bool_or(confirmed) as confirmed
+	FROM measurement
+	JOIN input ON input.input_no = measurement.input_no
+	JOIN report ON report.report_no = measurement.report_no
+	WHERE test_start_time >= current_date - interval '2 months'
+	AND test_start_time < current_date - interval '1 day'
+	AND test_name = 'web_connectivity'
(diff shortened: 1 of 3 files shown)


af/shovel/centrifugation.py
```diff
@@ -1685,6 +1685,44 @@ def update_explorer_metrics(pgconn, bucket):
...
+	DISTINCT input as input,
+	test_start_time,
+	probe_cc,
+	probe_asn,
+	bool_or(confirmed) as confirmed
+	FROM measurement
+	JOIN input ON input.input_no = measurement.input_no
+	JOIN report ON report.report_no = measurement.report_no
+	WHERE test_start_time >= current_date - interval '2 day'
+	AND test_start_time < current_date - interval '1 day'
```
hellais: It is not necessary that the `current_date` is going to be the time at which this is executed. It's better to use the `bucket_date` constraint to INSERT.

af/shovel/centrifugation.py
```diff
@@ -1685,6 +1685,51 @@ def update_explorer_metrics(pgconn, bucket):
...
+    JOIN autoclaved ON autoclaved.autoclaved_no = report.autoclaved_no
+	WHERE bucket_date = %s
+	AND test_name = 'web_connectivity'
+	GROUP BY input, test_start_time, probe_cc, probe_asn, bucket_date
+) as wc
+GROUP BY test_day, probe_cc, probe_asn, bucket_date;
+ON CONFLICT (test_day, probe_cc, probe_asn, bucket_date) DO
+UPDATE
+SET msm_count = EXCLUDE.msm_count,
+    confirmed_count = EXCLUDE.confirmed_count
```
hellais: `<bikeshed>`I guess we should re-indent this`</bikeshed>`

af/shovel/centrifugation.py
```diff
@@ -1685,6 +1685,51 @@ def update_explorer_metrics(pgconn, bucket):
 UPDATE
 SET count = EXCLUDED.count;''', [bucket])
 
+        # See:  016-ooexpl_wc_confirmed.install.sql
+        update_ooexpl_wc_confirmed = '''
+INSERT INTO ooexpl_wc_confirmed
```
hellais, line 1690: I think the issue is that we need to pass here the names of the columns

hellais: It should say something like:
```
INSERT INTO ooexpl_wc_confirmed (confirmed_count, msm_count, test_day, bucket_date, probe_cc, probe_asn)
```

hellais:
This looks good to me, I think we should deploy and merge (or merge and deploy).

PR #205, opened 2019-06-21
Add analysis tool

FedericoCeratto:
Generates database metrics published at:
https://mon.ooni.nu/grafana/d/FNVKM2-Wk/api-and-amsdb?orgId=1&from=now-7d&to=now

Related to https://github.com/ooni/sysadmin/pull/326


af/analysis/analysis.py
```diff
@@ -0,0 +1,879 @@
...
+from sqlalchemy import create_engine
+import pandas as pd
+import psycopg2
+import matplotlib
+
+matplotlib.use("Agg")
+import matplotlib.pyplot as plt
+import seaborn as sns
+
+from metrics import setup_metrics
```
hellais: I suggest you use the relative import syntax here to avoid confusing it with the module `metrics`.

That is doing this:
```
from .metrics import setup_metrics
```

af/analysis/analysis.py
```diff
@@ -0,0 +1,879 @@
...
+@metrics.timer("populate_countries")
+def populate_countries():
+    ## Used only once to create a persistent list of countries
+    dbengine.execute(
+        """
+        CREATE UNLOGGED TABLE countries (
+            probe_cc CHARACTER(2) NOT NULL,
+            msm_count BIGINT NOT NULL
+        );
+        CREATE INDEX ON countries (msm_count);
```
hellais: Why are you creating an index on the count of measurements for the country?

You may also want to look at the `ooexpl_bucket_msm_count` table which has similar information in it already.

hellais:
@FedericoCeratto is this PR still relevant?

hellais:
This is relevant to: https://github.com/ooni/pipeline/issues/256

hellais:
I will review this in this sprint.

hellais (approved):
I did a cursory review of this and it looks good to me.

I say we merge it!

PR #207, opened 2019-07-02
Match fingerprints by country and globally

FedericoCeratto:
Currently the fingerprint matching ignores the country code.


bassosimone (approved):
lgtm :whale:

PR #208, opened 2019-07-04
Add pipeline metrics

FedericoCeratto:
Add metric support to centrifugation
Improve test_integration.py: support configurable docker image tag, delete old Postgres container
Test if docker pings
Add docstrings
General refactor
Add tox.ini

### DIFF
--- a/af/shovel/Dockerfile
+++ b/af/shovel/Dockerfile
@@ -5,7 +5,7 @@ FROM ubuntu:16.04
 # All the actions are done within single RUN to reduce number of layers & overall fetch size.
 RUN set -ex \
     && apt-get update \
-    && apt-get install -y python2.7 liblz4-tool python-ujson python-yaml python-psycopg2 python-six \
+    && apt-get install -y python2.7 liblz4-tool python-ujson python-yaml python-psycopg2 python-six python-statsd \
     && : \
 # Whole python-lz4 repo is clonned as `git` can't do shallow clone of single
 # SHA1 commit and github tarball makes setuptools-scm unhappy while building.
--- a/af/shovel/centrifugation.py
+++ b/af/shovel/centrifugation.py
@@ -36,6 +36,7 @@ from oonipl.utils import dclass
 
 from oonipl.metrics import setup_metrics
 
+# Send metrics to the docker0 interface ipaddr
 metrics = setup_metrics(host="172.17.0.1", name="centrifugation")
 
 # It does NOT take into account metadata tables right now:
@@ -175,6 +176,7 @@ def dns_ttl(ttl):
 ########################################################################
 
 
+@metrics.timer("load_autoclaved_index")
 def load_autoclaved_index(autoclaved_index):
     """
     Loads the autoclaved index file and return the list of compressed
@@ -268,6 +270,7 @@ def create_temp_table(pgconn, table, definition):
         c.execute(create_table.format(table=table, definition=definition))
 
 
+@metrics.timer("copy_meta_from_index")
 def copy_meta_from_index(pgconn, ingest, reingest, autoclaved_index, bucket):
     # Writing directly to `report` table is impossible as part of `report`
     # table is parsed from data files, same is true for `measurement` table.
@@ -616,6 +619,7 @@ def load_global_duplicate_reports(pgconn):
         DUPLICATE_REPORTS.update(_[0] for _ in c)
 
 
+@metrics.timer("delete_data_to_reprocess")
 def delete_data_to_reprocess(pgconn, bucket):
     # Everything in *_meta is either ingested from scratch or re-ingested.
     # Some other rows in SOME tables (depending on code_ver) should also be deleted.
@@ -786,6 +790,7 @@ def prepare_destination(pgconn, stconn, bucket_code_ver, bucket_date):
     return report, msm, msm_exc, badmeta, ver_feeders, sink_list, feeder_list
 
 
+@metrics.timer("copy_data_from_autoclaved")
 def copy_data_from_autoclaved(pgconn, stconn, in_root, bucket, bucket_code_ver):
     TrappedException = None if FLAG_FAIL_FAST else Exception
 
@@ -923,6 +928,7 @@ def iter_autoclaved_datum(pgconn, autoclaved_root, bucket):
                                 "Short LZ4 frame", filename, frame_off, intra_off
                             )
                         datum = ujson.loads(datum)
+                        metrics.incr('iter_autoclaved_datum')
                         yield code_ver, autoclaved_no, report_no, msm_no, datum
                 for _ in iter(functools.partial(fd.read, 4096), ""):
                     pass  # skip till EOF
@@ -931,6 +937,8 @@ def iter_autoclaved_datum(pgconn, autoclaved_root, bucket):
                 print "Processed {}: {:.1f} MiB, {:.1f} s, {:.1f} MiB/s".format(
                     filename, size_mib, real, size_mib / real
                 )
+                metrics.incr("processed_data_MB", size_mib)
+                metrics.gauge("processing_data_speed_MBps", size_mib / real)
                 db_cksum = (
                     file_size,
                     file_crc32,
@@ -2225,6 +2233,7 @@ DATA_TABLES = (
 )
 
 
+@metrics.timer("update_explorer_metrics")
 def update_explorer_metrics(pgconn, bucket):
     with pgconn.cursor() as c:
         c.execute(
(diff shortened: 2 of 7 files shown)


af/shovel/Dockerfile
```diff
@@ -5,7 +5,7 @@ FROM ubuntu:16.04
 # All the actions are done within single RUN to reduce number of layers & overall fetch size.
 RUN set -ex \
     && apt-get update \
-    && apt-get install -y python2.7 liblz4-tool python-ujson python-yaml python-psycopg2 python-six \
+    && apt-get install -y python2.7 liblz4-tool python-ujson python-yaml python-psycopg2 python-six python-statsd\
```
bassosimone: Nit: maybe keep a whitespace before the `\`

af/shovel/centrifugation.py
```diff
@@ -34,6 +34,10 @@
 from oonipl.sim import sim_shi4_mm3_layout, sim_shi4_mm3_text
 from oonipl.utils import dclass
 
+from oonipl.metrics import setup_metrics
+
+metrics = setup_metrics(host="172.17.0.1", name="centrifugation")
```
bassosimone, line 40: I believe we should probably document what this IP address is. I cannot figure out immediately what is its purpose, but I believe the `172.17` space is used by Docker, right?

af/shovel/oonipl/metrics.py
```diff
@@ -0,0 +1,24 @@
...
+"""
+
+from os.path import basename, splitext
+
+import statsd
+
+
+def setup_metrics(host="localhost", name=None):
+    """Setup metric generation. Use dotted namespaces e.g.
+    "pipeline.centrifugation"
```
bassosimone, line 14: Maybe you can `assert` or `throw` if a dotted namespace is important?

tests/requirements.txt
```diff
@@ -1,3 +1,4 @@
+# Required by tests/test_integration.py
```
bassosimone, line 1: 👍

tests/test_integration.py
```diff
@@ -14,39 +21,54 @@
...
+        running[0].stop()
+        running[0].remove()
+
+    print("Starting Postgres container")
+    pg_container = docker_client.containers.run(
+        "postgres:9.6",
+        name=METADB_NAME,
+        hostname=METADB_NAME,
+        environment={"POSTGRES_USER": METADB_PG_USER},
+        ports={"5432/tcp": 25432},
```
bassosimone, line 44: This is the equivalent of `docker run --publish ...` right?

tests/test_integration.py
```diff
@@ -14,39 +21,54 @@
...
+        detach=True,
     )
     while True:
```
bassosimone, line 48: Meta thought: do we care about this potentially running forever?

tests/test_integration.py
```diff
@@ -14,39 +21,54 @@
...
 
+
 def pg_install_tables(container):
-    _, socket = container.exec_run(cmd="psql -U {}".format(METADB_PG_USER), stdin=True, socket=True)
-    if hasattr(socket, '_sock'):
+    _, socket = container.exec_run(
+        cmd="psql -U {}".format(METADB_PG_USER), stdin=True, socket=True
+    )
+    if hasattr(socket, "_sock"):
         socket = socket._sock
```
bassosimone, line 71: Why is this socket unwrapping necessary? (I think you probably didn't write this code but I am curious to know why accessing `_sock` is necessary in this context and maybe you know that.)

tests/test_integration.py
```diff
@@ -55,90 +77,100 @@ def pg_install_tables(container):
...
-                'bind': '/mnt',
-                'mode': 'rw'
-            }
-        },
+        mem_limit="512m",
+        user="{}:{}".format(os.getuid(), os.getgid()),
+        working_dir="/mnt",
+        links={METADB_NAME: None},
+        volumes={REPO_ROOT: {"bind": "/mnt", "mode": "rw"}},
         privileged=True,
```
bassosimone: Do you know why this container needs to be `--privileged`?

tests/test_integration.py
```diff
@@ -55,90 +77,100 @@ def pg_install_tables(container):
...
+        sys.stdout.write(line.decode("utf8"))
         sys.stdout.flush()
     print("runtime: {}".format(time.time() - start_time))
     return shovel_container
 
+
 def pg_conn():
-    return psycopg2.connect('host={} user={} port={}'.format('localhost', METADB_PG_USER, 25432))
+    return psycopg2.connect(
+        "host={} user={} port={}".format("localhost", METADB_PG_USER, 25432)
```
bassosimone: I'd consider whether we can avoid hardcoding the port number

bassosimone (changes requested):
I was able to understand the diff and it mostly looks good to me. I have asked a number of clarifying questions and provided some suggestions. The most pressing concern at the moment is that the CI is failing with multiple errors. My understanding is that somehow `statsd` is missing.

af/shovel/oonipl/metrics.py
FedericoCeratto: The dotted structure is for our convenience to create a hierarchy of metrics. If a component is clearly identified by a single word we can use it without dots.

FedericoCeratto:
Example of metrics (displayed by Netdata). The ingestion speed is the most interesting:

image
image

af/shovel/Dockerfile
FedericoCeratto: Done

af/shovel/centrifugation.py
FedericoCeratto: Updated

tests/test_integration.py
FedericoCeratto: It hasn't been a problem so far but a better try/finally to stop it even on test failure would be good.

tests/test_integration.py
FedericoCeratto: It appears only in 2 places and this is just a test harness...

tests/test_integration.py
FedericoCeratto: Unsure. It seems to work without it.

tests/test_integration.py
FedericoCeratto: Unsure. It was added in ecc43915a70027cb096465e802ba3360ceadf0a1 "Make the tests work with python2"

tests/test_integration.py
FedericoCeratto: It starts the container on a local port but there might be some differences.

bassosimone: Okay, thanks

tests/test_integration.py
bassosimone: Maybe we can add a TODO comment saying that we can make this change in the future?

tests/test_integration.py
bassosimone: I think it makes sense to document this as a py2 quirk with a comment?

tests/test_integration.py
bassosimone: I suggest to mention this in a comment as a possible future improvement?

bassosimone (approved):
Now CI is working. I have left a couple more follow up comments for possible quick documentation changes that we may want to make. :whale:

tests/test_integration.py
hellais: This is the relevant issue: https://github.com/docker/docker-py/issues/2255#issuecomment-475270012

bassosimone: Thanks! Can we perhaps add a comment in the code mentioning the issue?

tests/test_integration.py
hellais: Added a comment.

tests/test_integration.py
hellais: Added a comment for this.

hellais (approved):
I took a look at the diff and it looks good to me.

I committed a couple of comments directly to this branch. 

The CI is passing so for me this is good to merge.

@FedericoCeratto what was the reason for using statsd for collecting metrics vs using prometheus?

tests/test_integration.py
bassosimone: @hellais can we also take note of this?

hellais: I filed an issue for it here: https://github.com/ooni/pipeline/issues/211.

hellais:
> what was the reason for using statsd for collecting metrics vs using prometheus?

To preserve knowledge writing here what @FedericoCeratto told me privately. Basically we use Statd because it's quicker and eaiser to immediately see the metrics and gagues while developing which makes the dev cycle faster.
These are getting exposed locally using [netdata](https://github.com/netdata/netdata).

hellais (approved):
This looks good to me. Thanks for working on it and I can only imagine the pain doing that rebase!

PR #213, opened 2019-08-15
Implement fastpath

FedericoCeratto:
Implements ooni/spec/issues/145
Requirements and features:

- [x] Fetch data from:
  - [x] Database + raw cans on S3 for historical data
  - [x] Collectors: reports from up to 24 hours
  - [x] Collectors: reports fetched in real time
- [x] Convert YAML reports to JSON
- [x] Apply fingerprinting with global/country/ISP/local granularity
- [x] Apply scoring on web_connectivity test using simple anomaly heuristics
- [x] On-disk file caching
- [x] Packaging and sandboxing
- [x] SQL table generation
- [ ] SQL table deletion
- [x] Basic unit and functional testing
- [ ] Extensive unit and functional testing


af/fastpath/fastpath/core.py
```diff
@@ -0,0 +1,574 @@
...
+        log.setLevel(logging.DEBUG)
+
+    conf.conffile = root / "etc/fastpath.conf"
+    log.info("Using conf file %r", conf.conffile)
+    conf.vardir = root / "var/lib/fastpath"
+    conf.cachedir = conf.vardir / "cache"
+    conf.s3cachedir = conf.cachedir / "s3"
+    conf.dfdir = conf.vardir / "dataframes"
+    conf.outdir = conf.vardir / "output"
+    conf.msmtdir = conf.outdir / "measurements"
```
hellais, line 99: `pathlib.Path` is pretty neat,  I didn't know of this.

af/fastpath/fastpath/core.py
```diff
@@ -0,0 +1,574 @@
...
+        if not f.is_file():
+            continue
+        age_s = now - f.stat().st_atime
+        if age_s > threshold:
+            log.debug("Deleting %s", f)
+            metrics.gauge("deleted_cache_file_age", age_s)
+            # TODO: delete
+
+
+expected_colnames = {
```
hellais, line 138: I don't see where this variable is being used.

af/fastpath/fastpath/core.py
```diff
@@ -0,0 +1,574 @@
...
+    "whatsapp_endpoints_dns_inconsistent",
+    "whatsapp_endpoints_status",
+    "whatsapp_web_failure",
+    "whatsapp_web_status",
+}
+
+
+
+
+def unroll_requests_in_measurement(measurement):
```
hellais: I don't see this used either

af/fastpath/fastpath/core.py
```diff
@@ -0,0 +1,574 @@
...
+    """Generate a trivial id of the measurement to allow upsert if needed
+    - 32-bytes hexdigest
+    - Deterministic / stateless with no DB interaction
+    - Malicious/bugged msmts with collisions on report_id/input/test_name lead
+    to different hash values avoiding the collision
+    - Malicious/duplicated msmts that are semantically identical to the "real"
+    one lead to harmless collisions
+    """
+    # NOTE: we want the id to stay the same when a msmt is fetched from SSH
+    # and from a can on ooni-data-private/canned
```
hellais, line 408: I believe this will be the case, though beware that if instead of using canned data you use autoclaved data the hash of the JSON will not match, because some extra keys are added in autoclaving (for example the `bucket_date`).

af/fastpath/fastpath/core.py
```diff
@@ -0,0 +1,574 @@
...
+        with metrics.timer("full_run"):
+            try:
+                msm_jstr, measurement = msm_tup
+                if measurement is None:
+                    measurement = ujson.loads(msm_jstr)
+                msm_jstr, tid = trivial_id(measurement)
+                fn = generate_filename(tid)
+                writeout_measurement(msm_jstr, fn, conf.update)
+                matches = match_fingerprints(measurement)
+                summary = score_measurement(measurement, matches)
```
hellais, line 479: These two functions are applied to every measurements no matter the type. I think you should consider having some logic in place where the processing logic applied to measurements is dependent on the `test_name` key value. For example you probably don't want to be doing the `match_fingerprints` for middlebox test results and the `score_measurement` logic for `web_connectivity` and `facebook_messenger` will be different.

af/fastpath/fastpath/core.py
```diff
@@ -0,0 +1,574 @@
...
+            prefix = fp["header_prefix"]
+            if name in headers and headers[name].startswith(prefix):
+                matches.append(fp)
+                log.debug("matched header prefix %s %r", msm_cc, prefix)
+
+    return matches
+
+
+@metrics.timer("score_measurement")
+def score_measurement(msm, matches):
```
hellais, line 350: Maybe this should be called `score_web_connectivity`

af/fastpath/fastpath/core.py
```diff
@@ -0,0 +1,574 @@
...
+        yield measurement_tup
+
+
+@metrics.timer("match_fingerprints")
+def match_fingerprints(measurement):
+    """Match fingerprints against HTTP headers and bodies.
+    """
+    # TODO: apply only on web_connectivity
+    msm_cc = measurement["probe_cc"]
+    if msm_cc not in fingerprints:
```
hellais: My understanding is that you use `ZZ` in the fingerprints dict to indicate ANY country. In the case in which there aren't country specific fingerprints maybe we still want to run the fingerprint check against the fingerprints of `ZZ` and I think this check will skip that.

af/fastpath/fastpath/core.py
hellais: Looking at the code more I see that for `.yaml` measurements you are calling `normalize_entry` and passing it a fixed `bucket_date` or `2018-05-07` (maybe this value should be more clearly something hardcoded like a non-existent bucket_date) and the filename is set to `??`.

af/fastpath/fingerprints.csv
```diff
@@ -0,0 +1,58 @@
...
+49,US,,Via,,1.1 MLD-C-Barracuda.mld.org (http_scan_byf/3.5.16)
+50,US,,Via,,1.1 forcepoint-wcg.chambersburg.localnet
+51,US,,Location,http://filter.esu9.org:8080/webadmin/deny/index.php,
+52,US,,Location,http://ibossreporter.edutech.org/block/bp.html,
+53,US,,Location,http://alert.scansafe.net/alert/process,
+54,US,,Location,http://184.168.221.96:6080/php/urlblock.php,
+55,US,,Location,https://gateway.wifast.com:443/wifidog/login/,
+56,US,,Location,https://mpswebfilterwashbu.mpls.k12.mn.us:6082/php/uid.php,
+57,PA,,Via,,1.1 barracuda.tropigas.com.pa (http_scan_byf/3.3.1)
+58,BE,,Via,,1.1 webfilter.stjohns.net (http_scan_byf/3.5.16)
```
hellais: Is this a duplicate of the fingeprints inside of `utils.py` in the dict?

af/fastpath/fastpath/normalize.py
```diff
@@ -0,0 +1,811 @@
...
+        except Exception as exc:
+            yield normalize_entry(entry, "2018-05-07", "??", esha)
+
+
+def iter_yaml_lz4_reports_normalized(fn):
+    """Yields normalized measurements from a yaml.lz4 file
+    """
+    for r in iter_yaml_lz4_reports(fn):
+        off, entry_len, esha, entry, exc = r
+        yield normalize_entry(entry, "2018-05-07", "??", esha)
```
hellais: Perhaps these magic numbers `2018-05-07` and `??` should be defined as a constant variable and set to a value which is more clearly something hardcoded true (for example `2010-01-02` before there were any OONI measurements).

af/fastpath/fastpath/normalize.py
```diff
@@ -0,0 +1,811 @@
...
+    # TODO: tests these in
+    # test_normalize_yaml_sanitise_tcp_connect_bridge_reach
+    if entry["test_name"] == "tcp_connect":
+        entry = sanitise_tcp_connect(entry, BRIDGE_DB)
+
+    elif entry["test_name"] == "bridge_reachability":
+        entry = sanitise_bridge_reachability(entry, BRIDGE_DB)
+
+    return entry
+
```
hellais, line 485: It looks like a lot of the code in here comes from `daily_workflow.py`. I think we should find a way to avoid having this code be duplicated. Ideally both the fastpath and the batch pipeline would share the same module for doing this type of normalisation or at least some part of this code should be shared by both components.

hellais (commented):
I did a first pass at reading all the code and ran it in devel mode on the fastpath host.

Overall the architecture looks solid.

I left several questions and comments inline with the code.

af/fastpath/fastpath/core.py
FedericoCeratto: It even has convenient functions like Path(...).write_text(...)

af/fastpath/fastpath/core.py
FedericoCeratto: The idea is to extend it to handle any msmt.

af/fastpath/fastpath/core.py
FedericoCeratto: It's not used at the moment. Added comment.

af/fastpath/fastpath/core.py
FedericoCeratto: Removed

af/fastpath/fastpath/core.py
FedericoCeratto: There was a TODO in setup_fingerprints for this. Updated to handle ZZ.

af/fastpath/fingerprints.csv
FedericoCeratto: Yes, dropped.

af/fastpath/fastpath/normalize.py
FedericoCeratto: In the long term, yes. I plan to have this code well tested, merged and run and then backport it.

af/fastpath/fastpath/core.py
FedericoCeratto: Updated in the last commit. Added testing and temporarily disabled handling YAML.

af/fastpath/fastpath/normalize.py
FedericoCeratto: See comment above.

af/fastpath/fastpath/core.py
FedericoCeratto: The idea is to apply them to all msmts for consistency. Then, inside each function we'll implement more heuristics over time to handle each msmt type. Worst case we simply return early.

af/fastpath/fastpath/tests/test_functional.py
```diff
@@ -121,32 +115,119 @@ def notest_core(cans):
...
+            assert hash(entry) == expected[n]
+
+
+# def test_normalize_yaml(cans):
+#    # data_format_version is not 0.2.0
+#    can = "testdata/2013-05-05/20130505T103213Z-VN-AS24173-http_requests-no_report_id-0.1.0-probe.yaml.lz4"
+#    #fn = "fastpath/tests/data/20190912T115043Z-ndt7-20190912T115043Z_AS15169_0B9KifXvlmX0yttm3Q5rfUL9YYed8uXrvRtPELmZuuFek7NJDf-AS15169-US-probe-0.2.0.yaml"
+#    for r in iter_yaml_lz4_reports(can):
+#        off, entry_len, esha, entry, exc = r
+#        entry = normalize_entry(entry, "2018-05-07", "??", esha)
```
hellais: Why is this commented out?

af/fastpath/fastpath/tests/test_functional.py
```diff
@@ -121,32 +115,119 @@ def notest_core(cans):
...
+
+
+## YAML normalization
+
+def FIXME_test_normalize_yaml_2016(cans):
```
hellais: What is failing here to warrant the FIXME keyword?

af/fastpath/fastpath/tests/test_functional.py
FedericoCeratto: Moved to tests/test_functional_normalize.py with a code cleanup.
Normalization is still not stable and temporarily disabled. To be enabled in future after the MVP is out.

af/fastpath/fastpath/tests/test_functional.py
FedericoCeratto: (see comment above)

hellais:
As discussed with @FedericoCeratto we are going to land this in master 🎉

PR #215, opened 2019-08-16
Add assert to check if confirmed is > 0

hellais:
Somewhere after #207 the confirmed flag stopped getting computed properly.


hellais:
LGTM

PR #216, opened 2019-08-19
Fixture based testing

hellais:
This adds support for using single report files for instrumenting testing.

As part of adding test we identified and fixed several bugs.

### DIFF
--- a/af/shovel/oonipl/pg.py
+++ b/af/shovel/oonipl/pg.py
@@ -214,7 +214,7 @@ def pg_binquote(s):
 def pg_uniquote(s):
     if isinstance(s, str):
         s = unicode(s, "utf-8")
-    assert isinstance(s, unicode)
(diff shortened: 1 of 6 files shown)


af/shovel/oonipl/pg.py
```diff
@@ -214,7 +214,7 @@ def pg_binquote(s):
 def pg_uniquote(s):
     if isinstance(s, str):
         s = unicode(s, "utf-8")
-    assert isinstance(s, unicode)
+    assert isinstance(s, unicode), "{} not unicode".format(s)
```
FedericoCeratto: `"%r is not unicode" % s`   to show the repr of s

FedericoCeratto (approved):
(Discussed on Slack)

af/shovel/oonipl/pg.py
hellais: Ok

PR #220, opened 2019-08-30
Use Tox from Travis

FedericoCeratto:
Use Tox from Travis CI for consistency. tox-travis will also allow testing with py3 in future.


hellais:
LGTM

PR #228, opened 2019-10-01
Add event detector

FedericoCeratto:
Add event detector and RSS feed generation
Implements #229
Related to #213 and https://github.com/ooni/spec/issues/144


af/fastpath/debian/detector.service
```diff
@@ -0,0 +1,39 @@
...
+# Hardening
+CapabilityBoundingSet=CAP_SETUID CAP_SETGID
+SystemCallFilter=~@clock @debug @cpu-emulation @keyring @module @mount @obsolete @raw-io @reboot @swap
+NoNewPrivileges=yes
+PrivateDevices=yes
+PrivateTmp=yes
+ProtectHome=yes
+#ProtectHome=read-only
+ProtectSystem=strict
+ProtectSystem=full
```
bassosimone: Is `ProtectSystem=full` overriding or augmenting `ProtectSystem=strict`?

af/fastpath/fastpath/detector.py
```diff
@@ -0,0 +1,520 @@
...
+    FROM measurement
+    JOIN report ON report.report_no = measurement.report_no
+    JOIN input ON input.input_no = measurement.input_no
+    WHERE measurement_start_time >= %(start_date)s
+    AND probe_cc = %(cc)s
+    AND test_name = %(test_name)s
+    AND input = %(inp)s
+
+    ORDER BY measurement_start_time
+    """
```
bassosimone: I assume your end goal is to use this function to implement also the one above?

af/fastpath/debian/control
```diff
@@ -14,6 +14,7 @@ Depends: ${misc:Depends},
  ${python3:Depends},
  python3-paramiko,
  python3-statsd,
- python3-systemd
+ python3-systemd,
+ nginx
```
bassosimone, line 31: Is this redundant, considering that we are using Ansible to deploy hosts?

bassosimone (commented):
Hello @FedericoCeratto, here are my preliminary comments!

I like the principle of using Debian functionality as much as possible! I am unfortunately not familiar enough yet with our Ansible infrastructure, so I've asked a question regarding that.

I believe it would be beneficial to factor the bits where we perform measurement analysis in a different module with unit tests, so it would be easier (at least for me) to understand and visualize what is the typical input and the typical output. (That is, I'm not advocating for full coverage, even though this is nice, but rather to have unit tests as live examples of building blocks functionality.)

Thank you!

af/fastpath/debian/detector.service
FedericoCeratto: It's overriding it, I should remove the other.
https://www.freedesktop.org/software/systemd/man/systemd.exec.html#Sandboxing

hellais:
I am a bit concerned of the approach being so much reliant on debian packages for deploying stuff.

I fear this may lead to bits of our infrastructure using a certain pattern and others using a different one, making it necessary to know more bits of the system in order to do a new deployment.

I guess that if you publish the debian packages to some location and then the ansible playbook fetches a prebuilt packages, that is ok, though I would like to understand better what is the reason to go away from the current ansible based infrastructure.

I think it's easier to reason about a system (and make cross-cutting changes) when the patterns used in all the infrastructure are few and consistent.

I don't think it's necessary to use Docker (though it does make our life significantly easier in many regards), but I would like to better understand what direction you are thinking of going by introducing debian packaging in here.

af/fastpath/fastpath/detector.py
FedericoCeratto: Those are 2 different use cases: filtering by test_name/cc/input or getting everything.

af/fastpath/debian/control
FedericoCeratto: The dependencies are also used to run devel version on workstations,  run unit/integ tests and ensure that we are really deploying the required deps at runtime

hellais (approved):
We further discussed the deb based deployed and I think we have found a solution which can work.

For me this branch is good to be landed in master.

PR #231, opened 2019-10-05
Fastpath requirements

hellais:
These are some small changes that I have had to make to the fastpath to be able to run it in dev mode on macOS.

### DIFF
--- a/af/fastpath/fastpath/portable_queue.py
+++ b/af/fastpath/fastpath/portable_queue.py
@@ -1,71 +1,35 @@
-import six
-import platform
-import multiprocessing
-from multiprocessing.queues import Queue as BaseQueue
+# -*- coding: utf-8 -*-
+"""
+Portable multiprocessing queue singleton
+"""
 
-# taken from: https://github.com/binux/pyspider/blob/d49605d6d1f07d325ff7a4ca311dc5405ed93159/pyspider/libs/multiprocessing_queue.py
+import multiprocessing as mp
 
-class SharedCounter(object):
-    """ A synchronized shared counter.
-    The locking done by multiprocessing.Value ensures that only a single
-    process or thread may read or write the in-memory ctypes object. However,
-    in order to do n += 1, Python performs a read followed by a write, so a
-    second process may read the old value before the new one is written by the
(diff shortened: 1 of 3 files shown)


af/fastpath/fastpath/portable_queue.py
```diff
@@ -0,0 +1,71 @@
...
+    This class comes almost entirely from Eli Bendersky's blog:
+    http://eli.thegreenplace.net/2012/01/04/shared-counter-with-pythons-multiprocessing/
+    """
+
+    def __init__(self, n=0):
+        self.count = multiprocessing.Value('i', n)
+
+    def increment(self, n=1):
+        """ Increment the counter by n (default = 1) """
+        with self.count.get_lock():
```
FedericoCeratto: A whole dedicated class to call get_lock() seems a bit bloated. I would do the locking in the Queue class. Also it would allow decrementing the value directly instead of using "increment(-1)"

af/fastpath/fastpath/portable_queue.py
```diff
@@ -0,0 +1,71 @@
...
+            self.count.value += n
+
+    @property
+    def value(self):
+        """ Return the value of the counter """
+        return self.count.value
+
+
+class MultiProcessingQueue(BaseQueue):
```
FedericoCeratto: Wrapping a Queue instead of subclassing it sounds safer: we avoid touching the internals and doing the super() calls and we implement only the methods that are needed by the fastpath. The same wrapper can be used on every OS so that we always exercise the same code. The cost of using a lock should be negligible.

af/fastpath/fastpath/portable_queue.py
hellais: Yeah I guess so. I figured I would steal some already tested and written code though rather than writing it myself, but I am also ok with simplifying it.

af/fastpath/fastpath/portable_queue.py
hellais: I wasn't really sure how you were using the Queue and what plans you had for it, so it seemed safer to have an equivalent portable Queue. It's actually pretty crazy that the standard library python Queue is not portable even just between macOS and linux.

hellais:
Looks good to me!

PR #233, opened 2019-10-05
Fastpath platform

hellais:
As a way for me to familiarise myself with the fastpath I added support for extracting the `platform` annotation key which tells us which platform generated a specific measurement.

I needed to this also for a report and I was able to do it using the fastpath which was fun!

This PR is based on top of my other PR which implement fixes requires for me to run this on my macOS system (see: https://github.com/ooni/pipeline/pull/231).


FedericoCeratto:
Discussed in confcall - LGTM

PR #242, opened 2019-10-30
Ingest new metadata in fastpath for blocking detection

FedericoCeratto:
Version 0.3
Detect and score blocking of:
- Facebook Messenger
- Telegram
- Whatsapp
- Header field manipulation test
- Vanilla Tor connectivity test
- TCP connect test
- DASH Dynamic Adaptive Streaming over HTTP
- Meek

Basic handling (without scoring) of:
- Psiphon
- NDT

Minor improvements

### DIFF
--- a/af/fastpath/fastpath/core.py
+++ b/af/fastpath/fastpath/core.py
@@ -355,6 +355,52 @@ def match_fingerprints(measurement):
 
     return matches
 
+@metrics.timer("score_measurement_facebook_messenger")
+def score_measurement_facebook_messenger(msm, summary):
+    tk = msm["test_keys"]
+    del ms
+
+    # TODO recompute all these keys in the pipeline
+    # If the value of these keys is false (inconsistent) there is something
+    # fishy
+    consistency_keys = [
+        'facebook_b_api_dns_consistent',
+        'facebook_b_api_reachable',
+        'facebook_b_graph_dns_consistent',
+        'facebook_b_graph_reachable',
+        'facebook_edge_dns_consistent',
+        'facebook_edge_reachable',
+        'facebook_external_cdn_dns_consistent',
+        'facebook_external_cdn_reachable',
+        'facebook_scontent_cdn_dns_consistent',
+        'facebook_scontent_cdn_reachable',
+        'facebook_star_dns_consistent',
+        'facebook_star_reachable',
+        'facebook_stun_dns_consistent'
+    ]
+    # These are keys that if they are true it means there is something fishy
+    anomaly_keys = [
+        'facebook_tcp_blocking',
+        'facebook_dns_blocking'
+    ]
+
+    scores = {f"blocking_{l}": 0.0 for l in LOCALITY_VALS}
+    s = 0
+    for key in consistency_keys:
+        v = tk.get(key, None)
+        if v == False:
+            s += 0.5
+        scores[key] = v
+
+    for key in anomaly_keys:
+        v = tk.get(key, None)
+        if v == False:
+            s += 0.5
+        scores[key] = v
+    scores["blocking_general"] = s
+    summary["scores"] = scores
+    return summary
+
 
 @metrics.timer("score_measurement_telegram")
 def score_measurement_telegram(msm, summary):
@@ -519,6 +565,8 @@ def score_measurement(msm, matches):
 
     if msm["test_name"] == "telegram":
         return score_measurement_telegram(msm, summary)
+    if msm["test_name"] == "facebook_messenger":
+        return score_measurement_facebook_messenger(msm, summary)
 
     if msm["test_name"] == "whatsapp":
         return score_measurement_whatsapp(msm, summary)
--- a/af/fastpath/fastpath/core.py
+++ b/af/fastpath/fastpath/core.py
@@ -378,6 +378,17 @@ def falsekeys(d, keys):
     return True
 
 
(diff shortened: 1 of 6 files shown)


af/fastpath/fastpath/sshfeeder.py
```diff
@@ -78,7 +79,7 @@ def __init__(self, conf, hostname):
         self.sftp = ssh.open_sftp()
         # assert self.sftp.get_channel().get_transport()._preferred_compression[1] == 'zlib'
         self.new_downloads = []
-        self._archive_dir = f"/data/{self.hostname}/archive"
+        self._archive_dir = ARCHIVE_DIR
```
hellais: Does this imply reading only from the oonified collectors?

FedericoCeratto: Yes

FedericoCeratto:
Rebased on master

af/fastpath/fastpath/core.py
```diff
@@ -350,52 +351,599 @@ def match_fingerprints(measurement):
...
-        "input",
-        "measurement_start_time",
-        "probe_asn",
-        "probe_cc",
-        "report_id",
-        "test_name",
-        "test_start_time",
+    # TODO: use assertions for unknown bugs
+    rid = msm.get("report_id", "")
+    url = "https://explorer.ooni.org/measurement/{}".format(rid) if rid else "no rid"
```
hellais, line 398: This is a good idea.

You probably also want to check for `input` and if so put it in the URL too.

af/fastpath/fastpath/core.py
```diff
@@ -350,52 +351,599 @@ def match_fingerprints(measurement):
...
+    sname = msm.get("software_name", "unknown")
+    sversion = msm.get("software_version", "unknown")
+    if id > 0:
```
hellais, line 401: What is the logic by which the `id` is getting set/passed to this function?

af/fastpath/fastpath/core.py
```diff
@@ -350,52 +351,599 @@ def match_fingerprints(measurement):
...
+        keys = (keys,)
+    for k in keys:
+        if d.get(k, None) != True:
+            return False
+
+    return True
+
+
+def falsekeys(d, keys):
+    """Check for values set to True in a dict"""
```
hellais, line 371: `<bikeshed>`
Maybe these functions can have somewhere in them a the word `all`, maybe `all_keys_true`?

Initially I didn't understand it was looping over all of them and it's actually pretty similar to the python `all` construct.

af/fastpath/fastpath/core.py
hellais: I see you are always calling this with the value of `0` which should result in it going into the code path below. Why then not just check for `== 0`?

af/fastpath/fastpath/tests/test_functional.py
```diff
@@ -1,40 +1,74 @@
...
+# aws_secret_access_key = ...
+#
+# Explore bucket from CLI:
+# AWS_PROFILE=ooni-data-private aws s3 ls s3://ooni-data-private/canned/2019-07-16/
 
-@pytest.fixture
-def cans():
-    """
-    Download interesting cans from S3 to a local directory
+BUCKET_NAME = "ooni-data-private"
```
hellais, line 31: It's unfortunate we can't run these on the CI due to these files being in the private bucket. We should evaluate if we should open up the cans and move them to the public s3 bucket.

hellais (approved):
I took a look at the PR and it looks good to merge. Good work!

The comment I had were mostly to make me understand better the code base and some minor suggestions.

Feel free to merge this when you think it's ready.

af/fastpath/fastpath/core.py
FedericoCeratto: The idea here is to slowly build up a corpus of known bugs that we detect and ignore, and log out new ones.

af/fastpath/fastpath/core.py
FedericoCeratto: The id is just an unique identifier for the bug so we can correlate logs and code. I could have used the line number of the caller in the core.py but that would change over time.
The description field already contains an optional, longer description and even bug number on GitHub, so the id is just a small number.

af/fastpath/fastpath/core.py
FedericoCeratto: Good point, updated.

af/fastpath/fastpath/tests/test_functional.py
FedericoCeratto: As discussed before, the fastpath is publishing the msmts in real time on purpose it publishing the raw cans should be the same. However we could also have a dedicated s3 user configured in Travis CI

af/fastpath/fastpath/core.py
hellais: I figured it was going to be something like that, however it's unclear to me why you chose <= 0 as for **known** bugs and > 0 for unknown ones.

Perhaps you can have a single int value for all unknowns and everything else (which will be a range) for known bugs?

I would also mention how to use the `id` parameters in `logbug` docstring.

FedericoCeratto: I'm indeed using 0 for known bugs because the description replaces the id. The negative numbers are currently unused so we can use them in future if needed.

PR #245, opened 2019-11-15
Fastpath 0.6

FedericoCeratto:
Log to stdout when --stdout is passed or on OSX. Useful for manual runs.
Workaround for queue.close bug.
(to be rebased on fastpath-im that already contains v. 0.3)


hellais (approved):
This looks good to me!

PR #246, opened 2019-11-15
Fastpath env

hellais:
Based on the fastpath-stdout branch.

This is useful so I can run on local database instance a task to import just a selection of the data.

### DIFF
--- a/af/fastpath/fastpath/db.py
+++ b/af/fastpath/fastpath/db.py
@@ -29,6 +29,9 @@ DB_HOST = "hkgmetadb.infra.ooni.io"
 DB_USER = "shovel"
 DB_NAME = "metadb"
 DB_PASSWORD = "yEqgNr2eXvgG255iEBxVeP"  # This is already made public
+PG_DSN = f"host={DB_HOST} user={DB_USER} dbname={DB_NAME}"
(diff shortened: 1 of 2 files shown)


FedericoCeratto:
The detector uses a cli option --db-host . I'll implement the same semantic so they stay consistent.

af/fastpath/fastpath/db.py
```diff
@@ -40,9 +40,12 @@ def _ping():
...
 
 
-def setup():
+def setup(conf) -> None:
     global conn, _autocommit_conn
-    dsn = f"host={DB_HOST} user={DB_USER} dbname={DB_NAME} password={DB_PASSWORD}"
+    if conf.db_uri:
+        dsn = conf.db_uri
+    else:
+        dsn = f"host={DB_HOST} user={DB_USER} dbname={DB_NAME} password={DB_PASSWORD}"
```
hellais, line 48: Maybe this can be refactored as:
`<bikeshed>`
```
    dsn = f"host={DB_HOST} user={DB_USER} dbname={DB_NAME} password={DB_PASSWORD}"
    if conf.db_uri:
        dsn = conf.db_uri
```

PR #249, opened 2019-11-19
Improve metrics

FedericoCeratto:
Add metrics to measure generation of output files and the location of fingerprints matches in HTML bodies.


bassosimone (approved):
:whale:

PR #271, opened 2019-12-09
Fastpath 0.8: Add anomaly and confirmed columns

FedericoCeratto:
Add columns to ease integration with API and Explorer.

### DIFF
--- a/af/fastpath/fastpath/core.py
+++ b/af/fastpath/fastpath/core.py
@@ -1049,9 +1049,13 @@ def msm_processor(queue):
                     matches = []
                 confirmed = bool(len(matches))
                 scores = score_measurement(measurement, matches)
+                # Generate anomaly, confirmed and failure to keep consistency
+                # with the pipeline, allowing simple queries in the API and
+                # in manual analysis and keep compatibility with Explorer
                 anomaly = (scores.get("blocking_general", 0.0) > 0.5)
+                failure = scores.get("accuracy", 1.0) < 0.5
                 db.upsert_summary(
-                    measurement, scores, anomaly, confirmed, tid, fn, conf.update
+                    measurement, scores, anomaly, confirmed, failure, tid, fn, conf.update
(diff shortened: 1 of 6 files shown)


af/fastpath/fastpath/core.py
```diff
@@ -1046,8 +1047,12 @@ def msm_processor(queue):
                     matches = match_fingerprints(measurement)
                 else:
                     matches = []
+                confirmed = bool(len(matches))
```
hellais, line 1050: IIRC the fingerprints used by the fastpath differ from those inside of the batch pipeline.

Do we maybe want to mention somewhere this and add a TODO to re-align the two?

FedericoCeratto: I'll create a task in the project dashboard.

af/oometa/019-fastpath-confirmed.install.sql
```diff
@@ -0,0 +1,13 @@
+-- Add confirmed and anomaly columns to fastpath table
+-- Formatted with pgformatter 3.3
```
hellais, line 2: The rollback file here is missing.

af/fastpath/fastpath/core.py
hellais: Can you link it in here?

af/oometa/019-fastpath-confirmed.install.sql
FedericoCeratto: Added.

af/fastpath/fastpath/core.py
FedericoCeratto: Update: as discussed, with #278 we'll  keep the row in the fastpath db table and read from there. This make the re-align less urgent.

hellais (approved):
LGTM

FedericoCeratto:
Deployed

Issue #263, opened 2019-11-22
Discuss and do some exploration for data analysis needed by research reports


hellais:
This can also involve looking at:
https://github.com/ooni/pipeline/issues/61
https://github.com/ooni/pipeline/issues/45
https://github.com/ooni/pipeline/issues/46

and is a direct requirement to completing:
https://github.com/ooni/pipeline/issues/260
& https://github.com/ooni/pipeline/issues/261

FedericoCeratto:
I've spent few hours experimenting with:
1) Importing categories from input JSON files into the database
1.1) ...same, plus extracting the historical changes to the files using git history
2) modify the event detector webapp to add an "olap-like" query to slice data cubes

hellais:
We have created the citizenlab table which includes the category codes necessary for the creation of research reports.

This can be considered done in the previous cycle.

PR #274, opened 2019-12-12
Add domain_input and citizenlab tables and update tool

FedericoCeratto:
Implements https://github.com/ooni/api/issues/135 https://github.com/ooni/pipeline/issues/263 https://github.com/ooni/pipeline/issues/258


hellais (approved):
We discussed this and this looks good to me.

We should not forget, though, to add also the indexes which we created manually on the DB.

Merge this when you think it's most appropriate.

hellais (changes requested):
For the SQL migration script, the rollback file is missing.

hellais (approved):
🐳

FedericoCeratto:
Deployed 0.8~pr271-77

PR #276, opened 2019-12-28
Monitor database for new measurements

FedericoCeratto:
Implements https://github.com/ooni/backend/issues/219


FedericoCeratto:
https://github.com/ooni/pipeline/pull/205

af/analysis/analysis/analysis.py
```diff
@@ -942,11 +945,140 @@ def generate_slow_query_summary(conf):
...
+
+    Runs in a dedicated thread and writes in its own .prom file
+
+    This is the most important function, therefore it pings the SystemD watchdog
+    """
+    log.info("Started monitor_measurement_creation thread")
+    # TODO: switch to OOID
+
+    INTERVAL = 60 * 5  # half of the watchdog interval in the analysis.service file
+    nodeexp_path = "/run/nodeexp/db_metrics.prom"
```
hellais, line 969: Do we want to have this hardcoded in here?

af/analysis/analysis/analysis.py
```diff
@@ -942,11 +945,140 @@ def generate_slow_query_summary(conf):
...
+    if no_systemd == False:
+        watchdog.notify("READY=1")
+
+    cycle_seconds = 0
+
+    while True:
+        try:
+            log.info("Gathering fastpath count")
+            conn, dbengine = setup_database_connections(conf.standby)
+            delta = timedelta(minutes=5)
```
hellais, line 1039: Maybe `5` should be a variable which is in sync with `INTERVAL = 60 * 5`?

af/analysis/analysis/analysis.py
```diff
@@ -942,11 +945,140 @@ def generate_slow_query_summary(conf):
...
 
+@metrics.timer("monitor_measurement_creation")
+def monitor_measurement_creation(conf):
+    """Monitor measurements created by fastpath and traditional pipeline
+    to detect and alert on inconsistencies
+
+    Runs in a dedicated thread and writes in its own .prom file
+
+    This is the most important function, therefore it pings the SystemD watchdog
+    """
```
hellais, line 964: It's perhaps useful to add a comment on the logic of the queries that follow, as in that we are sampling every interval and comparing the number of measurements in the sample window period in the fastpath and the batch pipeline.

hellais (commented):
Left some review comments.

af/analysis/analysis/analysis.py
FedericoCeratto: Added

af/analysis/analysis/analysis.py
FedericoCeratto: Yes, it's close to where it's being used. monitor_measurement_creation() is meant to be self-contained

af/analysis/analysis/analysis.py
FedericoCeratto: No, this is independent from the run interval, e.g. we could run it every minute.

hellais (approved):
LGTM 🐳 !

PR #281, opened 2020-01-09
Fastpath 0.10 Add Tor and Psiphon enum values

FedericoCeratto:
Handle missing input msmt field
Implement scoring for Psiphon
Switch S3 bucket

### DIFF
--- a/af/fastpath/fastpath/core.py
+++ b/af/fastpath/fastpath/core.py
@@ -921,7 +921,28 @@ def score_psiphon(msm) -> dict:
     Returns a scores dict
     """
     scores = {f"blocking_{l}": 0.0 for l in LOCALITY_VALS}
-    tk = msm["test_keys"]
+    tk = msm.get("test_keys", {})
+
+    # https://github.com/ooni/spec/blob/master/nettests/ts-015-psiphon.md
+    failure = tk.get("failure", None)
+    bootstrap_time = tk.get("bootstrap_time", 0)
+
+    if failure is None:
+        if bootstrap_time == 0:
+            # should not happen
+            logbug(4, "invalid psiphon msmt", msm)
+            scores["accuracy"] = 0.0
+
+        else:
+            # success
+            scores["accuracy"] = 1.0
+
+    else:
+        # if bootstrap_time == 0: # there was an error bootstrapping Psiphon
+        # else: # there was an error when using Psiphon
+        scores["accuracy"] = 1.0
+        scores["blocking_general"] = 1.0
+
     if "resolver_ip" not in msm:
         logbug(0, "no resolver_ip", msm)
         scores["accuracy"] = 0.0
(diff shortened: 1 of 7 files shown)


FedericoCeratto:
Related to https://github.com/ooni/api/pull/158

FedericoCeratto:
Deployed at 82c6ce6b337296de99d3090b79bf67a91e39dde1

af/fastpath/fastpath/core.py
```diff
@@ -921,7 +921,28 @@ def score_psiphon(msm) -> dict:
...
+
+        else:
+            # success
+            scores["accuracy"] = 1.0
+
+    else:
+        # if bootstrap_time == 0: # there was an error bootstrapping Psiphon
+        # else: # there was an error when using Psiphon
+        scores["accuracy"] = 1.0
+        scores["blocking_general"] = 1.0
```
bassosimone, line 944: I think we want to distinguish the case where psiphon bootstrap from the one where it does not bootstrap. We may well do this in a followup PR; don't consider this blocking, just discuss if you disagree or keep track of it if you agree.

bassosimone (approved):
🐳 

let's just discuss, as mentioned, whether we want to consider bootstrap+error differently from bootstrap+success

af/fastpath/fastpath/tests/test_functional.py
```diff
@@ -23,20 +23,22 @@
...
+# [ooni-data]
 # aws_access_key_id = ...
 # aws_secret_access_key = ...
 #
 # Explore bucket from CLI:
 # AWS_PROFILE=ooni-data-private aws s3 ls s3://ooni-data-private/canned/2019-07-16/
 
-BUCKET_NAME = "ooni-data-private"
+# TODO: drop the boto3 code and use only s3feeder
+BUCKET_NAME = "ooni-data"
```
hellais, line 34: Since you are using the ooni-data bucket, you don't need any access keys anymore. You should look into making all boto requests using unsigned requests mode.

This can be done via:
```
>>> import boto3
>>> from botocore import UNSIGNED
>>> from botocore.config import Config
>>> s3 = boto3.client('s3', config=Config(signature_version=UNSIGNED))
```

hellais (approved):
Left a comment on switching to using unsigned boto requests.

This can either be done as part of this PR or documented as another issue and defered to future work.

I think it's important to do this though, as it would enable anybody to pickup the fastpath and start using it on their computer with zero configuration.

hellais:
Landing this to unblock: https://github.com/ooni/api/pull/158

af/fastpath/fastpath/tests/test_functional.py
hellais: Migrated this to: https://github.com/ooni/backend/issues/279

hellais (approved):
🐳

PR #284, opened 2020-01-17
Fastpath 0.11 Retry SSH connections on failure

FedericoCeratto:
Implements https://github.com/ooni/backend/issues/229

### DIFF
--- a/af/fastpath/debian/changelog
+++ b/af/fastpath/debian/changelog
@@ -1,3 +1,9 @@
+fastpath (0.11) UNRELEASED; urgency=medium
+
+  * Retry SSH connections on failure
+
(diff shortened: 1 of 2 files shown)


hellais (commented):
Is it correct to say that you are only retrying once on failure?

Do we maybe want to support retrying multiple times with some sort of backoff?

hellais (approved):
After discussing this privately, I have understood how this logic works.

Basically inside of https://github.com/ooni/pipeline/blob/136d4373c773a8f31cc5424fc90a94beadf609df/af/fastpath/fastpath/sshfeeder.py#L219 you will loop over all the sources and if it fails you will retry at the next loop, which will be, by default, in 1 second after having done all the sources.

This looks good to me!

🐳

FedericoCeratto:
Deployed

PR #286, opened 2020-01-20
0.13 Do not score title_match=False - bug #360

FedericoCeratto:


hellais:
This is related to ooni/explorer#360

PR #288, opened 2020-01-27
Add counters table updater

FedericoCeratto:
Implements https://github.com/ooni/backend/issues/292

### DIFF
--- a/af/analysis/analysis/counters_table_updater.py
+++ b/af/analysis/analysis/counters_table_updater.py
@@ -1,0 +1,139 @@
+"""
+Updates `counters` table
+
+Runs in a dedicated thread
+
+"""
+
+# TODO: regenerate data for the previous day once a day
+
+
+import logging
+import time
+
+import psycopg2
+
+from analysis.metrics import setup_metrics
+
+log = logging.getLogger("analysis")
+metrics = setup_metrics(name="analysis")
+
+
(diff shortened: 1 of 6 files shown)


af/analysis/analysis/counters_table_updater.py
```diff
@@ -0,0 +1,139 @@
...
+def counters_table_updater(conf):
+    """Thread entry point
+    """
+    while True:
+        try:
+            _update_counters_table(conf)
+        except Exception as e:
+            log.exception(e)
+
+        time.sleep(3600)
```
hellais, line 126: Do we maybe want to explicit the interval at which it updates as a constant variable at the start of the file?

hellais (approved):
I left a very minor comment, but otherwise it looks good. We should probably look into replacing the `ooexpl_` tables with things that draw from these tables instead to reduce duplication and present more realtime data in Explorer (taking from the fastpath too).

af/analysis/analysis/counters_table_updater.py
FedericoCeratto: Unless there are reasons to do otherwise, I prefer to leave 1-time-used constants in the place where they are used in order to encourage the reader to understand how the code works.

hellais: Ok fair enough.

Issue #192, opened 2019-05-05
Add malware fingerprint for KE


anadahz:
A possible malware fingerprint for Kenya.

Example measurement:
https://explorer.ooni.io/measurement/20180611T174527Z_AS33771_xGoXddliTIGLP3NJUBkEnEL1ukvMZKs7YvbB7RNFb3tW4OKZR7?input=http:%2F%2Fprotectionline.org%2F

tag:` http://159.89.232.4/alert`

hellais:
Judging from the redirect chain triggering a redirect to the malware site via HTTPS, I suspect this is done by having compromised the websites and then using it to deliver malware.

hellais:
I am going to make a new label for these types of cases, which I think need to be investigated separately and we should not add them as blockpages.

PR #289, opened 2020-01-29
Add fingerprints from `fingerprintdb`

hellais:
This implements: https://github.com/ooni/backend/issues/281

### DIFF
--- a/af/fastpath/fastpath/utils.py
+++ b/af/fastpath/fastpath/utils.py
@@ -170,10 +170,13 @@ fingerprints = {
     ],
     "FR": [
         {"body_match": 'xtpage = "page-blocage-terrorisme"', "locality": "country"},
+        # https://github.com/ooni/pipeline/issues/184
+        # https://explorer.ooni.org/measurement/20170213T143932Z_AS57584_U1gMCvIFLNWIKnSrpWTLrN9oewaHiT0DdKjLaT3MiMqnBk3J3U?input=http:%2F%2Fwww.23.org
         {
-            "header_full": "1.1 proxy2.rmc.local:80 (Cisco-WSA/9.1.1-074)",
-            "header_name": "Via",
-            "locality": "country",
+            #"header_full": "1.1 proxy2.rmc.local:80 (Cisco-WSA/9.1.1-074)",
+            #"header_name": "Via",
+            "body_match": "<title>Notification: Policy: URL Filtering</title>",
+            "locality": "local",
         },
     ],
     "GB": [
--- a/af/fastpath/fastpath/utils.py
+++ b/af/fastpath/fastpath/utils.py
@@ -180,31 +180,48 @@ fingerprints = {
         },
     ],
     "GB": [
+        # https://github.com/ooni/pipeline/issues/188
+        # https://explorer.ooni.org/measurement/20160401T061018Z_ehdZlDMzbUsYhYSGSWalmqKnqGIynfVHQhKvocrcKEmFeynXHd?input=Trans500.com
         {
             "header_name": "Location",
             "header_prefix": "http://blocked.nb.sky.com",
             "locality": "isp",
         },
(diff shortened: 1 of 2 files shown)


af/fastpath/fastpath/utils.py
```diff
@@ -141,10 +141,35 @@
...
     ],
     "IN": [
+        # https://github.com/ooni/pipeline/issues/25#issue-154919607
+        # https://explorer.ooni.org/measurement/7AEt2OwqdZUzFMzyZd3bFwnCwXp7FqYCezpaoBWdkBIfxtLtX84mXBZnlPLaTUqI?input=http:%2F%2Fthepiratebay.se
         {
             "body_match": "The page you have requested has been blocked",
             "locality": "country",
-        }
+        },
+        # TODO: maybe we would like to support defining a fingerprint by regexp?
```
hellais, line 285: cc @FedericoCeratto

hellais:
31 issues done, 12 more to go!

af/fastpath/fastpath/utils.py
FedericoCeratto: It would be pretty expensive (and can potentially introduce vulnerabilities). Perhaps if we do the matching only on anomalies we could rethink of it?

FedericoCeratto (approved):

af/fastpath/fastpath/utils.py
```diff
@@ -261,60 +612,95 @@
...
+        # https://explorer.ooni.org/measurement/20130506T043500Z_IBzthJbAAPsdFZQLPrjMbAwlELzGqtabqMatKpBqqBSWnUyQnA?input=http:%2F%2Fwww.sos-reporters.net
         {
             "header_name": "Location",
             "header_prefix": "http://ezxcess.antlabs.com/login/index.ant?url",
             "locality": "local",
         }
     ],
     "ZZ": [
-        {
-            "header_full": "Kerio Control Embedded Web Server",
```
FedericoCeratto, line 299: The Kerio product can be used in other countries, hence the ZZ.

af/fastpath/fastpath/utils.py
```diff
@@ -135,35 +269,126 @@
...
+        # https://explorer.ooni.org/measurement/3N5bdjWAdVjZubaIyAxCCAg0HiZYWfT1YLgz6cI0zRq1XTjHzBmGg49AbRxOGILi?input=http:%2F%2Fwarc.jalb.de
+        {
+            "header_full": "WebProxy/1.0 Pre-Alpha"
+            "header_name": "Server",
+            "locality": "local",
         }
     ],
+    # https://github.com/ooni/pipeline/issues/192
+    # https://explorer.ooni.org/measurement/20180611T174527Z_AS33771_xGoXddliTIGLP3NJUBkEnEL1ukvMZKs7YvbB7RNFb3tW4OKZR7?input=http:%2F%2Fprotectionline.org%2F
+    #"KE": [
```
FedericoCeratto, line 331: Is this meant to be commented out?

af/fastpath/fastpath/utils.py
hellais: Yes you are right. I think it would actually be enough to also have a rule for headers which is `header_contains`. This would allow us to add a rule that is say:

```
"header_contains": "(Cisco-WSA/",
```

So that we can match all the Cisco-WSA equipment

FedericoCeratto (approved):

af/fastpath/fastpath/utils.py
FedericoCeratto: +1

af/fastpath/fastpath/utils.py
hellais: Yes, check the github issue for details: https://github.com/ooni/pipeline/issues/192#issuecomment-586171982

af/fastpath/fastpath/utils.py
hellais: I think that applies to many other case too though. I put in the real country of the measurement because I assumed you had automatically put it in ZZ because the probe_cc in the measurement was set to `ZZ`, yet the `resolver_address` mapped to a CL network.

PR #296, opened 2020-02-18
Fastpath 0.22 score http_invalid_request_line

FedericoCeratto:
Implement http_invalid_request_line scoring and tests
Many of the msms keys are optional


af/fastpath/fastpath/core.py
```diff
@@ -655,6 +655,40 @@ def score_measurement_hhfm(msm):
...
+    """Calculate measurement scoring for http_invalid_request_line
+    """
+    # https://github.com/ooni/spec/blob/master/nettests/ts-007-http-invalid-request-line.md
+    tk = msm["test_keys"]
+    rid = msm["report_id"]
+    scores = {f"blocking_{l}": 0.0 for l in LOCALITY_VALS}
+    sent = tk.get("sent", [])
+    received = tk.get("received", [])
+
+    if not len(sent) and not len(received):
```
hellais, line 697: I would also add a check here to see if the received list looks something like:
```
"received": [
"",
"",
"",
""
]
```

This can also be a sign of the backend being down, though it may also be a sign of a middlebox, so perhaps you want to bump down the accuracy, but still mention it as an anomaly?

bassosimone (approved):
🐳

hellais:
@FedericoCeratto should we maybe merge this?

FedericoCeratto:
Merged with https://github.com/ooni/pipeline/pull/309

PR #297, opened 2020-02-20
FP 0.18 Implement WhatsApp parsing

FedericoCeratto:
Parse raw report and ignore checks from probe due to https://github.com/ooni/probe-engine/issues/341

### DIFF
--- a/af/fastpath/fastpath/core.py
+++ b/af/fastpath/fastpath/core.py
@@ -494,6 +494,26 @@ def score_measurement_facebook_messenger(msm):
     return scores
 
 
+def _extract_tcp_connect(tk):
+    # https://github.com/ooni/spec/blob/master/data-formats/df-005-tcpconnect.md
+    # NOTE: this is *NOT* ts-008-tcp-connect.md
+    # First the probe tests N TCP connections
+    tcp_connect = tk.get("tcp_connect", [])
+    accessible_endpoints = 0
+    unreachable_endpoints = 0
+    for entry in tcp_connect:
+        s = entry.get("status", {})
+        success = s.get("success", None)
+        if success is True:
+            accessible_endpoints += 1
+        elif success is False:
+            unreachable_endpoints += 1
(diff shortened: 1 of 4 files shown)


af/fastpath/fastpath/utils.py
```diff
@@ -498,7 +498,7 @@
         {
             "body_match": "<iframe src=\"http://subblock.mts.ru/api",
             "locality": "isp"
-        }
+        },
```
hellais, line 501: 👍 apologies for the typos, should  probably have run it through a linter 😬

hellais (approved):
🐳

af/fastpath/fastpath/core.py
```diff
@@ -494,6 +494,26 @@ def score_measurement_facebook_messenger(msm):
     return scores
 
 
+def _extract_tcp_connect(tk):
+    # https://github.com/ooni/spec/blob/master/data-formats/df-005-tcpconnect.md
+    # NOTE: this is *NOT* ts-008-tcp-connect.md
```
bassosimone, line 499: Yeah, I know, this is confusing :-(

bassosimone: Since we can rename test helpers without too much pain, maybe `df-005-tcpdial.md`?

bassosimone (approved):
🐳 looks good to me

af/fastpath/fastpath/core.py
hellais: We should eventually relocate all legacy tests to some special tree or mark them as deprecated in a more obvious way. I would not worry to much about the naming clash as the key names are in fact identical `test_name: tcp_connect` and `tcp_connect: []`.

I do agree it's confusing though :P

FedericoCeratto: The tricky part is in the key names indeed. Let's just keep the comments in the code as reminders.

PR #302, opened 2020-03-03
Analysis 0.8 Coalesce NULL inputs

FedericoCeratto:
used to compare entries in pipeline and fastpath


FedericoCeratto:
Deployed. It lowered the count in https://mon.ooni.nu/grafana/d/0NwR3eUWz/fastpath?orgId=1&from=now-24h&to=now as expected.

hellais (approved):
Yes this makes sense.

Perhaps it's useful to also add a comment on that line stating why we are doing the COALESCE?

PR #303, opened 2020-03-04
FP 0.21 Improve type checking and fix log_ingestion_delay

FedericoCeratto:
Tested successfully with:
    mypy --ignore-missing-imports fastpath/core.py

### DIFF
--- a/af/fastpath/fastpath/s3feeder.py
+++ b/af/fastpath/fastpath/s3feeder.py
@@ -144,6 +144,7 @@ def fetch_cans(s3, conf, files):
         _cb.count += bytes_count
         _cb.total_count += bytes_count
         metrics.gauge("s3_download_percentage", _cb.total_count / _cb.total_size * 100)
+        log.debug("s3_download_percentage %d", _cb.total_count / _cb.total_size * 100)
         try:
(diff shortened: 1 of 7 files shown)


FedericoCeratto:
Deployed.

af/fastpath/fastpath/s3feeder.py
```diff
@@ -154,11 +159,15 @@ def _cb(bytes_count):
         # TODO: handle missing file
         log.info("Downloading can %s size %d MB" % (fn, size / 1024 / 1024))
         diskf.parent.mkdir(parents=True, exist_ok=True)
+        tmpf = diskf.with_suffix(".s3tmp")
```
hellais, line 162: Do you think it makes sense to add something to clean these files up in the event of the process being restarted or killed?

FedericoCeratto: This is meant for manual runs and there is no automated cleanup of the cached files (given that the pipeline does not know if we want to use the cached files only once). At the moment I'm deleting the whole cache manually. In future we can add trimming old files and temp files.

hellais (approved):
LGTM

PR #304, opened 2020-03-06
analysis 0.10: Add counters table and updater

FedericoCeratto:
Based on https://github.com/ooni/pipeline/pull/305

### DIFF
--- a/af/analysis/analysis/counters_table_updater.py
+++ b/af/analysis/analysis/counters_table_updater.py
@@ -23,32 +23,49 @@ metrics = setup_metrics(name="analysis")
 def _populate_counters_table(cur):
     log.info("Populating counters table from historical data")
     sql = """
-    INSERT INTO counters (measurement_start_day, test_name, probe_cc, probe_asn, input, anomaly, confirmed, failure, count)
+    INSERT INTO counters (measurement_start_day, test_name, probe_cc, probe_asn, input, anomaly_count, confirmed_count, failure_count, measurement_count)
     SELECT
-        date_trunc('day', measurement_start_time) AS measurement_start_day,
+        measurement_start_day,
         test_name::text,
         probe_cc,
         probe_asn,
         input,
-        anomaly,
-        confirmed,
-        measurement.exc IS NOT NULL AS failure,
-        COUNT(*) AS count
-    FROM
-        measurement
-        JOIN input ON input.input_no = measurement.input_no
-        JOIN report ON report.report_no = measurement.report_no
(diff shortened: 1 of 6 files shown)


af/analysis/analysis/counters_table_updater.py
```diff
@@ -0,0 +1,173 @@
...
+@metrics.timer("update_counters_table")
+def _update_counters_table(conf):
+    log.info("CTU: Started update_counters_table thread")
+    conn = connect_db(conf.active)
+    cur = conn.cursor()
+    if _table_is_empty(cur):
+        _populate_counters_table(cur)
+
+    log.info("CTU: Deleting today's data")
+    sql = "DELETE FROM counters WHERE measurement_start_day = CURRENT_DATE"
```
hellais, line 95: I think there is an issue with this pattern of assuming that you will not update the tables of data pertaining to a date in the past once it's past the CURRENT_DATE.

The problem is that:
1. The clocks of probes are not necessarily accurate
2. A measurement may be uploaded many days after it has been run

For these reason I believe you need to somehow also store in the columns the `bucket_date` or some value which is related to when the measurement was first seen in the fastpath/pipeline.

hellais: In the `ooexpl_` tables we use `bucket_date` as such key, but you probably can't use that as-is because you are doing this in the fastpath.

FedericoCeratto: Until we introduce a timestamp on the collector we'll have that ambiguity.
Essentially we would like to have:
 1) updates faster than once a day (aka from fastpath)
 2) stable data for past days that does not change over time (e.g. a chart from yesterday should not change)
 3) All measurements from any given day based on measurement_start_time accounted precisely

We can only pick 2 out of 3. 
There's a TODO at the top of the file to regenerate the previous day.

hellais: > Until we introduce a timestamp on the collector we'll have that ambiguity.


Yes I agree with this. Though collection time is not necessarily the same as measurement time.

You can run a measurement today and upload it in 2 weeks from now. This can happen for many reasons, maybe the OONI backend was done, maybe it was censored or maybe you simply wanted to save bandwidth while you were on a limited mobile link.

I would say that 2 is a requirement we can drop.

Since clients are able to upload measurements from the past, it is always going to be the case that charts may change in the event data.

FedericoCeratto:
Rebased on master; switched to unlogged table.

PR #305, opened 2020-03-13
analysis 0.9: Fix replication delay

FedericoCeratto:
The replication delay can be misleading in 2 ways due to the measurement based on `now() - pg_last_xact_replay_timestamp()`. The timestamp is based on the last write on the DB.
1) When the active and standby DBs are in sync it shows a delay > 0
2) When the active and standby DBs are out of sync it shows the time difference `now <--> last write on standby` instead of `last write on active <--> last write on standby`

This change fixes 1)


FedericoCeratto:
Related to https://github.com/ooni/sysadmin/issues/433

hellais (approved):
LGTM!

PR #312, opened 2020-04-24
Prio 0.1

FedericoCeratto:
Second iteration on URL prioritization MVP
Implements https://github.com/ooni/backend/issues/361


hellais (approved):
I took a look at the prioritisation algorithm and it looks good to me! 🐳

PR #313, opened 2020-04-28
Documentation generator

FedericoCeratto:


build_docs.ini
```diff
@@ -0,0 +1,137 @@
...
+  }
+  footer {
+    background-color: rgb(0, 54, 91);
+    padding: 2rem;
+    color: #fff;
+    font-size: 14px;
+    font-family: "Fira Sans", sans-serif;
+    margin-top: 2rem;
+    position: fixed;
+    bottom: 0;
```
sarathms: Maybe we don't need a fixed footer.

```suggestion
```

PR #315, opened 2020-05-01
FP 0.27 Stream cans: enable parallel can fetching and processing

FedericoCeratto:
Add flags to prevent writing to db and to disk
Generate ETA metric
Fix generator type hints
Add unit test
General code cleanup


hellais (approved):
This is very cool! Looks good to me!

FedericoCeratto:
Deployed version 0.27

PR #319, opened 2020-06-01
Analysis 0.15 add coverage

FedericoCeratto:
Also add configurable port number

### DIFF
--- a/af/analysis/analysis/analysis.py
+++ b/af/analysis/analysis/analysis.py
@@ -67,6 +67,7 @@ import psycopg2  # debdeps: python3-psycopg2
 from psycopg2.extras import RealDictCursor
 
 import matplotlib  # debdeps: python3-matplotlib
+
 matplotlib.use("Agg")
 import matplotlib.pyplot as plt
 import seaborn as sns  # debdeps: python3-seaborn
@@ -93,11 +94,19 @@ metrics = setup_metrics(name="analysis")
 node_exporter_path = "/run/nodeexp/analysis.prom"
 
 
+def setup_database_connection(c):
+    return psycopg2.connect(
+        dbname=c["dbname"],
+        user=c["dbuser"],
+        host=c["dbhost"],
+        password=c["dbpassword"],
+        port=c.get("dbport", 5432),
+    )
+
+
 @contextmanager
 def database_connection(c):
-    conn = psycopg2.connect(
-        dbname=c["dbname"], user=c["dbuser"], host=c["dbhost"], password=c["dbpassword"]
-    )
+    conn = setup_database_connection(c)
     try:
         yield conn
     finally:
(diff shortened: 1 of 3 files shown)


af/analysis/analysis/analysis.py
```diff
@@ -535,6 +555,100 @@ def summarize_core_density():
...
+        SELECT platform, probe_cc, count(*)
+        FROM (
+          SELECT platform, probe_cc
+          FROM fastpath
+          WHERE (
+              (probe_cc, input) IN (SELECT UPPER(cc), url FROM citizenlab)
+              OR
+              input IN (SELECT url FROM citizenlab WHERE cc = 'ZZ')
+            )
+            AND measurement_start_time > NOW() - interval '1 days'
```
hellais, line 603: Since the coverage_generator is run every 24h and it's applying the calculation on a 1 day interval from the current time, depending on when it's started for the first time it might run before the batch pipeline has processed the daily batch.

Perhaps the calculation should be done on a 24h window which is 1 day in the past?

hellais:
So these charts are looking at coverage by checking:

1. How many measurements we have in total from a given platform in a 60 day window
2. What percentage of the citizenlab test list is being covered on a certain day by country and platform

I think this is a great start in being able to quantify and assess coverage, but I can see a few limitations.

For **1** 

I think we should also be counting the number of unique networks and country codes being covered by each platform. The reason to do this is that I think that the absolute measurement count only tells one part of the picture and is very much impacted by the scheduling logic used by that particular platform. Since we currently don't have automated testing on the next generation probes, OONI Probe series 2.x will be contributing too much to these counts.

For **2**

I think looking at the % of coverage of URLs in the citizenlab list is only one piece of the puzzle. In light of the fact that we are working on prioritising URLs, we are in fact optimizing for not having complete or equal coverage of all URLs in the test lists, so I think we should think of some metric which takes this into account.

I think something which looks at some set of high priority URLs over time could be a way to go about dealing with this.
Maybe also using the outputs of the confirmed/anomaly analysis could also be a way to go. At the end of the day what we care about is ensuring that blocked sites are being detected and adequately measured, so I think that is something we should consider in this analysis engine too.

hellais (approved):
I left a comment explaining what I think could be done as future work and areas for improvements, that said I think this is good to go!

af/analysis/analysis/analysis.py
FedericoCeratto: I plan to change the run frequency in future while keeping the same daily window to monitor improvements when we deploy the URL prioritization.  NOW() makes it flexible.

FedericoCeratto:
0.15~pr319-282 deployed

FedericoCeratto:
0.15~pr319-283 deployed

FedericoCeratto:
[Comments discussed in backend meeting]
Merging this PR and letting the code bake in. Will get back to it in a week or two.

PR #322, opened 2020-07-10
Clickhouse

FedericoCeratto:


af/fastpath/fastpath/db.py
```diff
@@ -148,3 +156,141 @@ def upsert_summary(
...
+        "platform" String,
+        "filename" String,
+        "scores" String,
+        "anomaly" Int8,
+        "confirmed" Int8,
+        "msm_failure" Int8,
+        "control_failure" String,
+        "is_ssl_expected" Int8,
+        "page_len" Int32,
+        "page_len_ratio" Float32,
```
hellais, line 190: Another feature you might want is also HTTP response headers. You should at least return the boolean if they match.

af/fastpath/fastpath/db.py
```diff
@@ -148,3 +156,141 @@ def upsert_summary(
...
+        "server_cc" String,
```
hellais, line 191: It would also be useful to have the `blocking` determination done by the probe as a column on which to do feature extraction.

Other neat columns are:
`http_experiment_failure`
`dns_experiment_failure`

hellais (approved):
Left some small comments on improvements. Overall looks good to me. Merge when you see fit! 🐳

hellais:
This is superseeded by #371

PR #323, opened 2020-07-15
fastpath 0.30 switch to ams-pg.ooni.nu

FedericoCeratto:


hellais (approved):
This PR looks good to merge for me.

We should land it together with a fix that updates the address of the batch pipeline too.

FedericoCeratto:
Moving DB URI to configuration file

PR #327, opened 2020-11-10
Fastpath: various changes

hellais:
I start it using: `./run_fastpath --devel`

But then I get stuck at:

```
196 14746 INFO fastpath Using conf file /Users/art/code/ooni/pipeline/af/fastpath/etc/ooni/fastpath.conf
196 14746 INFO fastpath collectors: ['localhost']
197 14746 INFO fastpath Starting
199 14746 ERROR fastpath cannot pickle 'module' object
Traceback (most recent call last):
  File "/Users/art/code/ooni/pipeline/af/fastpath/fastpath/core.py", line 1288, in core
    [t.start() for t in workers]
  File "/Users/art/code/ooni/pipeline/af/fastpath/fastpath/core.py", line 1288, in <listcomp>
    [t.start() for t in workers]
  File "/Users/art/.pyenv/versions/3.8.6/lib/python3.8/multiprocessing/process.py", line 121, in start
    self._popen = self._Popen(self)
  File "/Users/art/.pyenv/versions/3.8.6/lib/python3.8/multiprocessing/context.py", line 224, in _Popen
    return _default_context.get_context().Process._Popen(process_obj)
  File "/Users/art/.pyenv/versions/3.8.6/lib/python3.8/multiprocessing/context.py", line 284, in _Popen
    return Popen(process_obj)
  File "/Users/art/.pyenv/versions/3.8.6/lib/python3.8/multiprocessing/popen_spawn_posix.py", line 32, in __init__
    super().__init__(process_obj)
  File "/Users/art/.pyenv/versions/3.8.6/lib/python3.8/multiprocessing/popen_fork.py", line 19, in __init__
    self._launch(process_obj)
  File "/Users/art/.pyenv/versions/3.8.6/lib/python3.8/multiprocessing/popen_spawn_posix.py", line 47, in _launch
    reduction.dump(process_obj, fp)
  File "/Users/art/.pyenv/versions/3.8.6/lib/python3.8/multiprocessing/reduction.py", line 60, in dump
    ForkingPickler(file, protocol).dump(obj)
TypeError: cannot pickle 'module' object
201 14746 INFO fastpath Shutting down workers
1201 14746 INFO fastpath Shutting down workers
2205 14746 INFO fastpath Join
Traceback (most recent call last):
  File "./run_fastpath", line 3, in <module>
    main()
  File "/Users/art/code/ooni/pipeline/af/fastpath/fastpath/core.py", line 1339, in main
    core()
[...]

PR #328, opened 2020-11-11
Fix bug in pipeline handling of http_invalid_request_line tests

### LINKED ISSUE #1278: probe-legacy: hirl: possibly invalid tampering logic
While inspecting [a OONI measurement](https://explorer.ooni.org/measurement/20201011T235114Z_AS12880_fNB7tEDBoOzAuDGS04aQ2ZLnYVHsKLDORc0BjIa7QALwTUYPpC) with @sarathms, we noticed that the JSON contained signs of tampering but the it also claimed that `tampering` was `false`. I inspected [the source code of hirl](https://github.com/ooni/probe-legacy/blob/master/ooni/nettests/manipulation/http_invalid_request_line.py).

My understanding of the source code is that the `test_` functions are scheduled to run either in parallel or in sequence. In any case, each of them registers a callback that calls `check_for_manipulation`.

It also seems to me that `check_for_manipulation` overwrites the value of `report['tampering']` unconditionally.

This means that, if, say, the first subtest to terminate founds tampering and the subsequent one does not find tampering, then the code will override the value of `tampering` to be `false`.

This seems to me a good explanation of the phenomenon observed by me and @sarathms in the above mentioned measurement. We should probably double check this by running a specific measurement where Jafar is used to block one of the cases measured by hhfm, so to confirm this hypothesis.

hellais:
Fixes: https://github.com/ooni/probe/issues/1278

### DIFF
--- a/af/fastpath/fastpath/core.py
+++ b/af/fastpath/fastpath/core.py
@@ -666,11 +666,12 @@ def score_http_invalid_request_line(msm):
         if s != r:
             anomaly = True
 
-    tampering = tk.get("tampering", False)
-    if tampering != anomaly:
-        scores["accuracy"] = 0.0
-        logbug(6, "Incorrect tampering flag", msm)
-        return scores
+    # We ignore the tampering flag due to: https://github.com/ooni/probe/issues/1278
+    # tampering = tk.get("tampering", False)
+    # if tampering != anomaly:
+    #     scores["accuracy"] = 0.0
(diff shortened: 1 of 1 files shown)


bassosimone (approved):
Thank you for fixing this, @hellais! I think we should go ahead 🐳 notwithstanding the errors in the integration tests because IIUC these failing tests are for the batch pipeline?

hellais:
Yes these tests are in fact disabled in my #327

PR #344, opened 2021-04-29
FP 0.38 add Signal scoring

### LINKED ISSUE #500: Fastpath: add Signal scoring

FedericoCeratto:
Fixes https://github.com/ooni/backend/issues/500

### DIFF
--- a/af/fastpath/fastpath/core.py
+++ b/af/fastpath/fastpath/core.py
@@ -1217,14 +1217,19 @@ def score_signal(msm) -> dict:
     """Calculates measurement scoring for Signal test
     Returns a scores dict
     """
+    # https://github.com/ooni/spec/blob/master/nettests/ts-029-signal.md
     scores = {f"blocking_{l}": 0.0 for l in LOCALITY_VALS}
     tk = msm.get("test_keys", {})
     if tk.get("failed_operation", True) or tk.get("failure", True):
         scores["accuracy"] = 0.0
 
-    for c in tk.get("tcp_connect", []):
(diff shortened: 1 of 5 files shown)


af/fastpath/fastpath/core.py
```diff
@@ -1213,6 +1213,22 @@ def score_dns_consistency(msm) -> dict:
...
+    scores = {f"blocking_{l}": 0.0 for l in LOCALITY_VALS}
+    tk = msm.get("test_keys", {})
+    if tk.get("failed_operation", True) or tk.get("failure", True):
+        scores["accuracy"] = 0.0
+
+    for c in tk.get("tcp_connect", []):
+        if c.get("status", {}).get("success", False) == False:
+            scores["blocking_general"] = 1.0
+
+    return scores
```
hellais, line 1234: You should be reading the keys:
```
{
    "signal_backend_failure": "FAILURE STRING",
    "signal_backend_status": "blocked"
}
```
See: https://github.com/ooni/spec/blob/master/nettests/ts-029-signal.md

hellais (changes requested):
It should use the signal_backend_status and failure keys to determine the blocking, not just the values of the tcp_connect test results as the experiment is also checking if the TLS certificate is good.

af/fastpath/fastpath/core.py
FedericoCeratto: Done

hellais (approved):
LGTM

PR #352, opened 2021-06-01
Publish coverage

FedericoCeratto:


.github/workflows/test_fastpath.yaml
```diff
@@ -20,3 +20,9 @@ jobs:
           cd af/fastpath
           PYTHONPATH=. pytest-3 -v --junitxml=junit/test-results.xml --cov=fastpath --cov-report=xml --cov-report=html \
             --cov-report=term fastpath/tests/test_unit.py fastpath/tests/test_functional.py
+
+      - name: Archive code coverage HTML pages
+        uses: actions/upload-artifact@v2
+        with:
+          name: coverage
+          path: af/fastpath/htmlcov
```
hellais, line 33: Maybe we want to use a third party service like coveralls for reporting coverage directly in the PR comments?

hellais: See: https://github.com/marketplace/actions/coveralls-github-action

FedericoCeratto: [discussed on Slack]

PR #355, opened 2021-06-07
Add support for DNS based blockpage fingerprints

hellais:


FedericoCeratto (approved):
We should also add an integration test later on.

PR #356, opened 2021-06-07
Add reprocessing tool

FedericoCeratto:


af/fastpath/fastpath/reprocessor.py
```diff
@@ -0,0 +1,322 @@
...
+def process_measurement(msm_tup, buf, conf, s3sig, db_conn):
+    """Process a msmt
+    If needed: create a new Entity tracking a jsonl file,
+      close and upload jsonl to S3 and upsert db
+    """
+    THRESHOLD = 20 * 1024 * 1024
+    msm_jstr, msm, _ = msm_tup
+    if msm is None:
+        msm = ujson.loads(msm_jstr)
+    # TODO: yaml?
```
hellais: From what I gather reading the code of `load_multiple` when the format is in YAML you should always be getting a non null msm value, so I guess there is no need to handle YAML normalization and parsing in here.
Or am I missing something?

FedericoCeratto: Yes, It's updated now.

hellais (approved):
LGTM

PR #359, opened 2021-07-02
0.24 Support SSL certs in test helper rotation

FedericoCeratto:


af/analysis/rotation.py
```diff
@@ -201,6 +241,136 @@ def drain_droplet_if_needed(db_conn, live_droplets, active_droplets_count):
...
+
+    cmd = [
+        "certbot",
+        "certonly",
+        "--dns-digitalocean",
+        "--dns-digitalocean-credentials",
+        certbot_creds,
+        "-d",
+        "'*.th.ooni.org'",
+    ]
```
hellais, line 268: Shouldn't this command be run in here somehow?

FedericoCeratto: It's commented out and there's a TODO for it

af/analysis/rotation.py
```diff
@@ -64,9 +79,25 @@
 
 conffile_path = "/etc/ooni/rotation.conf"
 setup_script_path = "/etc/ooni/rotation_setup.sh"
+nginx_conf = "/etc/ooni/rotation_nginx_conf"
```
hellais, line 82: Why not embed the nginx configuration content into this script itself? If not maybe it should be documented somewhere how this looks like.

hellais (approved):
Left a minor nitpick comment about documenting how the nginx configuration looks like. Otherwise looks good to merge.

af/analysis/rotation.py
FedericoCeratto: In order to store it in ansible.

PR #360, opened 2021-07-23
Add torsf test (pluggable transport)

FedericoCeratto:
Related to https://github.com/ooni/ooni.org/issues/771 https://github.com/ooni/ooni.org/issues/772
Implements https://github.com/ooni/ooni.org/issues/766

### DIFF
--- a/af/fastpath/fastpath/core.py
+++ b/af/fastpath/fastpath/core.py
@@ -1260,7 +1260,7 @@ def score_torsf(msm) -> dict:
     tk = msm.get("test_keys", {})
     failure = tk.get("failure")
     if failure:
-        scores["accuracy"] = 0.0
+        scores["blocking_general"] = 1.0
         return scores
 
     scores["extra"] = dict(bootstrap_time=tk.get("bootstrap_time"))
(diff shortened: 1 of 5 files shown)


bassosimone (approved):
🐳

af/fastpath/fastpath/core.py
```diff
@@ -1251,6 +1251,22 @@ def score_signal(msm) -> dict:
...
+def score_torsf(msm) -> dict:
+    """Calculate measurement scoring for Tor Snowflake
+    Returns a scores dict
+    """
+    # https://github.com/ooni/ooni.org/issues/772
+    scores = init_scores()
+    tk = msm.get("test_keys", {})
+    failure = tk.get("failure")
+    if failure:
+        scores["accuracy"] = 0.0
```
hellais: I don't think you should be setting the accuracy to zero in the case of `failure`. The meaning of the `failure` key is to indicate that there is an instance of blocking of the protocol.

FedericoCeratto:
"torsf" test name added to database

af/fastpath/fastpath/core.py
FedericoCeratto: Fixed

PR #364, opened 2021-09-14
0.43 Filter minican names

FedericoCeratto:

PR #365, opened 2021-09-15
Add support for extracting signal_backend_failure

hellais:


hellais:
Integrated it directly into: https://github.com/ooni/pipeline/pull/364

PR #369, opened 2021-11-11
0.47 Whatsapp: support http URLs

FedericoCeratto:


af/fastpath/fastpath/core.py
```diff
@@ -775,7 +775,8 @@ def score_measurement_whatsapp(msm):
             # has unexpected contents
             # Also note bug https://github.com/ooni/probe-legacy/issues/60
 
-        # TODO: handle elif url == "http://web.whatsapp.com/":
+        elif url == "http://web.whatsapp.com/":
+            webapp_accessible = b.get("failure", True) in (None, "", False)
```
bassosimone: I think here you should set `webapp_accessible` equals to False if:

1. there is a failure, or
2. the status code is not 302, or
3. the number of location headers is != 1, or
4. the location is not the HTTPS URL

hellais: I have implemented this via the last few commits.

@FedericoCeratto can you take a look at this?

af/fastpath/fastpath/core.py
```diff
@@ -696,6 +696,29 @@ def score_http_invalid_request_line(msm):
     return scores
 
 
+def get_http_header(resp, header_name, case_sensitive=False):
```
bassosimone, line 699: I _assume_ you are adding this `case_sensitive` switch because you anticipate it would be useful in the future to have this functionality in this function to perform comparison for tests that change the header case?

af/fastpath/fastpath/core.py
```diff
@@ -696,6 +696,29 @@ def score_http_invalid_request_line(msm):
...
+    if case_sensitive == False:
+        header_name = header_name.lower()
+
+    header_list = []
+
+    # backward compatibility with older measurements that don't have
+    # header_list
```
bassosimone, line 715: Thank you!

bassosimone (commented):
Thanks for taking the leadership on fixing this issue! The code looks good to me but I am left wondering why the function to compare headers has built-in support for lower-case comparison, which is disabled, when we're not using this functionality. I assume there may be a reason for doing that, but I don't immediately grasp it, so I felt it was proper to ask and clarify this aspect of the diff. Apart from this comment, I do not see any other issue in this diff related to its design or implementation. Thanks again! <3

af/fastpath/fastpath/core.py
hellais: Yeah, the idea behind this is that there might be cases in which we actually care to search for a header with case_sensitivity turned on (I am thinking for example of the middlebox tests where some header mangling may be performed by the middlebox).
Since it seemed easy to support this use case, I figured I would add a flag for it from the get-go.

Alternatively, if you feel that having unused functionality is not idea, we can also drop this and add it only when we actually need it.

bassosimone: > [...] the idea behind this is that there might be cases in which we actually care to search for a header with case_sensitivity turned on (I am thinking for example of the middlebox tests where some header mangling may be performed by the middlebox)

A-ha! This totally make sense, but it confused me because I did not think about it. Perhaps add some documentation to the function explaining why we have this support?

Thanks in general for adding support for the headers list, it makes me happy to see it!

bassosimone (approved):
🐳

af/fastpath/fastpath/core.py
```diff
@@ -696,6 +696,38 @@ def score_http_invalid_request_line(msm):
...
 
 
+def get_http_header(resp, header_name, case_sensitive=False):
+    """This function returns the HTTP header(s) matching the given
+       header_name from the headers. The returned value is a list
+       given that we may have multiple keys per header. If the new
+       headers_list field is present in the response, we'll use
+       that, otherwise we'll fallback to the headers map. We perform
+       case sensitive header names search by default. You can yet
+       optionally select case insenstive comparison, which is useful,
```
bassosimone:
```suggestion
       optionally select case insensitive comparison, which is useful,
```

PR #371, opened 2021-11-18
0.52 Support Clickhouse

FedericoCeratto:


af/fastpath/fastpath/core.py
```diff
@@ -1389,7 +1398,7 @@ def flag_measurements_with_wrong_date(msm: dict, msmt_uid: str, scores: dict) ->
     if delta < -3600:
         log.debug(f"Flagging measurement {msmt_uid} from the future")
         scores["accuracy"] = 0.0
-        scores["msg"] = "Measurement start time from the future"
+        scores["msg"] = "Incorrect measurement start time from the future"
```
hellais: Are we sure we want to change this msg string? I worry it's going to impact our ability to parse them by string matching if we have some of them marked with one string and others marked with the other.

af/fastpath/fastpath/db.py
```diff
@@ -156,3 +168,158 @@ def upsert_summary(
...
+        `measurement_uid` String,
+        `report_id` String,
+        `input` String,
+        `probe_cc` String,
+        `probe_asn` Int32,
+        `test_name` String,
+        `test_start_time` DateTime,
+        `measurement_start_time` DateTime,
+        `filename` String,
+        `scores` String,
```
hellais, line 199: Is the idea here to use the [clickhouse JSON functions](https://clickhouse.com/docs/en/sql-reference/functions/json-functions/) for interacting with the `scores` JSON string?

Did you consider the possibility of using a [Map](https://clickhouse.com/docs/en/sql-reference/data-types/map/) instead? If so why was the string data type chosen in place of Map?

af/fastpath/fastpath/db.py
```diff
@@ -156,3 +168,158 @@ def upsert_summary(
...
+        `software_name` String,
+        `software_version` String,
+        `control_failure` String,
+        `blocking_general` Float32,
+        `is_ssl_expected` Int8,
+        `page_len` Int32,
+        `page_len_ratio` Float32,
+        `server_cc` String,
+        `server_asn` Int8,
+        `server_as_name` String
```
hellais, line 214: Maybe since we are adding to the base schema fields for the future, we can also add here `test_runtime`?

af/fastpath/fastpath/db.py
```diff
@@ -156,3 +168,158 @@ def upsert_summary(
...
+    )
```
hellais, line 215: A more general consideration I have is that currently the additional fields which have been added are very web_connectivity specific and tests that are not web_connectivity will have these fields unset. I can imagine that in the future we are going to need to also implement some feature extraction that is specific for particular test types (ex. number of reachable directory authorities in the `tor` test) and I wonder how this is going to impact the evolution of the schema.
I can see one approach, which is the direction where we are going now, of adding everything to the fastpath table, but we do run the risk of the schema eventually becoming a bit bloated. Another approach could potentially be that of putting the test specific fields in dedicated tables and looking them up only when necessary via cross table joins.

af/fastpath/fastpath/core.py
FedericoCeratto: That was unwanted. Reversing.

af/fastpath/fastpath/db.py
FedericoCeratto: The JSON structure is nested.

af/analysis/analysis/citizenlab_test_lists_updater.py
```diff
@@ -107,36 +108,59 @@ def create_citizenlab_cc_idx(conn):
...
+def query_c(click, query: str, qparams: dict):
+    click.execute(query, qparams, types_check=True)
+
+
+@metrics.timer("rebuild_citizenlab_table_from_citizen_lab_lists")
+def update_citizenlab_table_click(conf, citizenlab):
+    """Overwrite citizenlab_flip and swap tables atomically"""
+    if conf.dry_run:
+        return
+    click = Clickhouse("localhost")
```
hellais, line 142: Shouldn't you be using `conf.clickhouse_host` here instead of hardcoding `localhost`?

af/analysis/analysis/citizenlab_test_lists_updater.py
```diff
@@ -107,36 +108,59 @@ def create_citizenlab_cc_idx(conn):
...
+    log.info("Emptying Clickhouse citizenlab_flip table")
+    q = "TRUNCATE TABLE citizenlab_flip"
```
hellais, line 144: How and where is the schema of the `citizenlab_flip` table defined?

af/analysis/analysis/citizenlab_test_lists_updater.py
```diff
@@ -107,36 +108,59 @@ def create_citizenlab_cc_idx(conn):
...
+    click.execute(q)
+
+    log.info("Inserting %d citizenlab table entries", len(citizenlab))
+    q = "INSERT INTO citizenlab_flip (domain, url, cc, category_code) VALUES"
+    click.execute(q, citizenlab, types_check=True)
+
+    log.info("Swapping Clickhouse citizenlab tables")
+    q = "EXCHANGE TABLES citizenlab_flip AND citizenlab"
+    click.execute(q)
```
hellais, line 153: Should you perhaps also truncate it at the end of this workflow so we don't keep duplicate data lying around in the DB?

af/fastpath/fastpath/core.py
hellais: I still see it in the PR, perhaps it still needs to be removed.

af/analysis/analysis/citizenlab_test_lists_updater.py
FedericoCeratto: No, it's designed to have only local connectivity as a safety feature.

af/fastpath/fastpath/core.py
FedericoCeratto: Reverted now

af/analysis/analysis/citizenlab_test_lists_updater.py
FedericoCeratto: A copy of citizenlab. Both new and old tables are kept around to help debugging. They are very small.

af/analysis/analysis/citizenlab_test_lists_updater.py
FedericoCeratto: It's done on purpose to be able to compare new and old data for debugging.

PR #372, opened 2021-11-30
0.51 Extract test runtime

FedericoCeratto:
Implements https://github.com/ooni/backend/issues/542

### DIFF
--- a/af/fastpath/fastpath/core.py
+++ b/af/fastpath/fastpath/core.py
@@ -1080,6 +1080,7 @@ def score_psiphon(msm) -> dict:
         logbug(0, "no resolver_ip", msm)
         scores["accuracy"] = 0.0
 
+    scores["extra"] = dict(test_runtime=msm.get("test_runtime"))
     return scores
 
 
@@ -1128,6 +1129,7 @@ def score_tor(msm) -> dict:
     else:
         scores["accuracy"] = 0.0
 
+    scores["extra"] = dict(test_runtime=msm.get("test_runtime"))
     return scores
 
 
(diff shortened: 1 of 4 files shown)


hellais (approved):
LGTM

af/fastpath/fastpath/core.py
```diff
@@ -1080,6 +1080,7 @@ def score_psiphon(msm) -> dict:
         logbug(0, "no resolver_ip", msm)
         scores["accuracy"] = 0.0
 
+    scores["extra"] = dict(test_runtime=msm.get("test_runtime"))
```
hellais: Psiphon also has the key `bootstrap_time` nested under `test_keys`. While we are at it, let's add support for extracting that as well.

af/fastpath/fastpath/core.py
```diff
@@ -1080,7 +1080,9 @@ def score_psiphon(msm) -> dict:
         logbug(0, "no resolver_ip", msm)
         scores["accuracy"] = 0.0
 
-    scores["extra"] = dict(test_runtime=msm.get("test_runtime"))
+    truntime = msm.get("test_runtime")
+    btime = tk.get("bootstrap_time")
```
hellais: Why don't you instead use the `bootstrap_time` variable that you already have available from above? I think it's more robust to use that since it's doing a `get` with a default value so it won't cause an exception if the key is missing.

FedericoCeratto: I though we assumed it's there and wanted to raise rather than set to 0. But we can instead set it to 0 and rely on accuracy being set to 0.

hellais (approved):
LGTM

FedericoCeratto:
0.51~pr372-24 deployed on test and fsn.
Deployed to prod.

PR #374, opened 2021-12-31
Add UZ blockpage fingerprints

vivivibo:


hellais:
Thanks for contributing to the OONI data processing pipeline with this new blockpage fingerprint!

af/fastpath/fastpath/utils.py
```diff
@@ -1629,6 +1629,20 @@
...
+	# https://explorer.ooni.org/measurement/20211231T141146Z_webconnectivity_UZ_8193_n1_Ap5yFzrVoL671ahx?input=http%3A%2F%2Fwww.europacasino.com%2F
+        {
+            "header_name": "Location",
+            "header_prefix": "http://reestr.mitc.uz/",
+            "locality": "country",
+        },
+	# https://explorer.ooni.org/measurement/20211231T172847Z_webconnectivity_UZ_8193_n1_kzei9uKT3oPETGzS?input=http%3A%2F%2Fikhwanonline.com%2F
+        {
+            "header_name": "Location",
+            "header_prefix": "http://reestr.mitc.uz/xtrm.html",
```
hellais: Since the matching is done based on the prefix, you only need to add the first rule as it will match in both cases.

I will edit the PR directly and merge it.

PR #380, opened 2022-02-25
0.57 Improve reprocessor

FedericoCeratto:
Support Clickhouse


af/fastpath/fastpath/core.py
```diff
@@ -1463,7 +1464,7 @@ def flag_measurements_with_wrong_date(msm: dict, msmt_uid: str, scores: dict) ->
         return
     try:
         recv_time = datetime.strptime(msmt_uid[:22], "%Y%m%d%H%M%S.%f_")
-        start_time = msm.get("measurement_start_time")
+        start_time = msm.get("measurement_start_time") or ""
```
hellais, line 1516: I guess the gotcha here is that we have using the `or ""` to convert a `None` `measurement_start_time` into a string so that the following call will fail with a `ValueError` instead of a `TypeError`.

af/fastpath/fastpath/reprocessor.py
```diff
@@ -163,55 +199,69 @@ def score_measurement_and_upsert_fastpath(msm, msmt_uid, do_update: bool) -> Non
...
+
+        rid = msm.get("report_id")
+        test_name = msm.get("test_name")
+        input_, domain = extract_input_domain(msm, test_name)
+        assert not isinstance(input_, list)
+        tn = test_name.replace("_", "")
+        cc = msm.get("probe_cc").upper()
+        desc = f"{msmt_uid} {tn} {cc} {rid} {input_}"
+    except Exception as e:
+        log.info(f"Ignoring broken measurement")
```
hellais: Can we add some additional logging information to identify which measurements are broken?

I think that once we do a run it would be useful to take a look at the broken ones and understand why we failed to load them.

I would say it's probably useful to record:
* Where it came from (the can or jsonl)
* The measurement_uid
* Increment a metric in statsd

af/fastpath/fastpath/normalize.py
```diff
@@ -575,21 +576,24 @@ def iter_yaml_msmt_normalized(data, bucket_tstamp: str, report_fn: str):
...
 
         if not entry:  # e.g. '---\nnull\n...\n'
             continue
         if "test_start_time" in entry and "test_start_time" in header:
             header.pop("test_start_time")
         entry.update(header)
         try:
-            yield normalize_entry(entry, bucket_tstamp, report_fn, esha)
+            d = normalize_entry(entry, bucket_tstamp, report_fn, esha_d)
+            msmt_uid = trivial_id(raw_entry, d)
```
hellais, line 637: Do we care to concatenate the entry header when computing the trivial_id?

af/fastpath/fastpath/s3feeder.py
```diff
@@ -66,23 +66,25 @@ def load_multiple(fn: str) -> Generator[MsmtTup, None, None]:
                 if m.name.endswith(".json"):
                     for line in k:
                         msm = ujson.loads(line)
-                        msmt_uid = trivial_id(msm)
+                        msmt_uid = trivial_id(line, msm)
+                        msm["measurement_uid"] = msmt_uid
```
hellais, line 75: Ok cool, so we are adding the `measurement_uid` to the output JSON entries. 

We should remember to update ooni/spec to reflect this improvement to the output data format.

af/fastpath/fastpath/db.py
```diff
@@ -52,14 +53,16 @@ def setup(conf) -> None:
...
+def extract_input_domain(msm: dict, test_name: str) -> tuple[str, str]:
     """Extract domain and handle special case meek_fronted_requests_test"""
-    input_ = msm.get("input", None)
+    input_ = msm.get("input") or ""
     if test_name == "meek_fronted_requests_test" and isinstance(input_, list):
-        domain = None if input_ is None else urlparse(input_[0]).netloc
-        input_ = ":".join(input_)
+        domain = input_[0]  # type: str
+        input_ = ",".join(input_)
+        input_ = "{" + input_ + "}"
```
hellais, line 62: Is there a reason to change the format of the input for the meek test results?

If we do implement this change, we should test if explorer is able to handle it properly. Otherwise we can maybe stick to the old format of separating the fronts with a `:`?

af/fastpath/fastpath/reprocessor.py
```diff
@@ -12,12 +12,14 @@
...
     jsonl/{testname}/{cc}/{ts}/00/{jsonlf.name}
   rows in the jsonl database table
-    upsert or insert-but-not-overwrite
+  rows in the fastpath database table
 
 Usage:
 export AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY
 PYTHONPATH=. ./fastpath/reprocessor.py ooni-data ooni-data-eu-fra-test --day 2015-1-1
 
+Note: the bundling of measurements into jsonl gz files has to remain deterministic
```
hellais, line 21: It might help to clarify more precisely what is meant by "deterministic" and why we need it.

Do we mean that the same measurements have to end up inside of the same filenames? That they end up in the same s3 paths? Or that the hash of the jsonl.gz files has to be exactly the same across runs?

af/fastpath/fastpath/core.py
FedericoCeratto: They are equally good at raising an exception but avoiding an optional None makes mypy happier

af/fastpath/fastpath/reprocessor.py
FedericoCeratto: Added "{desc}" and metric

af/fastpath/fastpath/normalize.py
FedericoCeratto: What's the benefit?

af/fastpath/fastpath/db.py
FedericoCeratto: It's what we already have in the database.

af/fastpath/fastpath/reprocessor.py
FedericoCeratto: It's what we discussed in the meetings and on Slack. Two runs of the reprocessor on different hosts or at different times will generate the same jsonl files.

af/fastpath/fastpath/db.py
hellais: Oh, I see now that we have in the database a mixture of both formats:
```
SELECT input
FROM fastpath
WHERE (test_name = 'meek_fronted_requests_test') AND (input LIKE '{%')
LIMIT 1

┌─input────────────────────────────────────────────┐
│ {a0.awsstatic.com,d2cly7j4zqgua7.cloudfront.net} │
└──────────────────────────────────────────────────┘
```

and
```
SELECT input
FROM fastpath
WHERE (test_name = 'meek_fronted_requests_test') AND (input NOT LIKE '{%')
LIMIT 1

┌─input──────────────────────────────────────────┐
│ a0.awsstatic.com:d2zfqthxsdq309.cloudfront.net │
└────────────────────────────────────────────────┘
```

hellais: So I tested the ones prefixed with `{` in explorer and while they don't look very pretty, they don't seem to break.

Screenshot 2022-03-03 at 18 15 38

I guess it's not super high priority to align these to the same format since it's a legacy test.

af/fastpath/fastpath/normalize.py
hellais: I guess that morally a YAML measurement is the header + the entry, but yeah you are probably right that there isn't any real benefit to doing so.

af/fastpath/fastpath/reprocessor.py
```diff
@@ -163,55 +199,71 @@ def score_measurement_and_upsert_fastpath(msm, msmt_uid, do_update: bool) -> Non
...
+
+        rid = msm.get("report_id")
+        test_name = msm.get("test_name")
+        input_, domain = extract_input_domain(msm, test_name)
+        assert not isinstance(input_, list)
+        tn = test_name.replace("_", "")
+        cc = msm.get("probe_cc").upper()
+        desc = f"{msmt_uid} {tn} {cc} {rid} {input_}"
+    except Exception as e:
+        log.info(f"Ignoring broken measurement {desc}")
```
hellais: Isn't the `desc` variable unbound in the exception handler?

af/fastpath/fastpath/core.py
```diff
@@ -377,6 +377,25 @@ def match_fingerprints(measurement):
...
 
 
+def g_or(d, key, default):
+    """Dict getter: return 'default' if the key is missing or value is None"""
+    # Note: 'd.get(key) or default' returns 'default' on any falsy value
+    v = d.get(key)
+    return default if v is None else v
+
+
+def g(d, *keys, default=None):
```
hellais, line 392: Maybe you should add a check to ensure len(keys) > 0, otherwise you will get an IndexError.

FedericoCeratto: If we forget to add  the keys throwing an IndexError would be ok.

af/fastpath/fastpath/normalize.py
hellais: We discussed this and it's probably not so helpful to have. Closing.

af/fastpath/fastpath/s3feeder.py
FedericoCeratto: This was based on a discussion on Slack. Now legacy msmts have the UID in them.

af/fastpath/fastpath/reprocessor.py
FedericoCeratto: Fixed.

af/fastpath/fastpath/tests/data/http_header_field_manipulation_2.json
```diff
@@ -0,0 +1 @@
+{"test_keys":{"requests":[],"agent":"agent","socksproxy":null},"test_start_time":"2019-10-28 11:56:47","input_hashes":[],"probe_ip":"127.0.0.1","id":"3282669b-2397-4d8f-bad3-9cd81825518e","test_helpers":{"backend":"http:\/\/37.218.247.95:80"},"probe_cc":"BR","test_runtime":1.3552720547,"input":null,"probe_city":null,"probe_asn":"AS28573","annotations":{"platform":"linux"},"software_name":"ooniprobe","software_version":"2.2.0","data_format_version":"0.2.0","report_filename":"2019-10-28\/20191028T115647Z-BR-AS28573-http_header_field_manipulation-20191028T115649Z_AS28573_eIrzDM4njwMjxBi0ODrerI5N03zM7qQoCvl4xpapTccdW0kCRg-0.2.0-probe.json","test_version":"0.2.0","bucket_date":"2019-10-28","test_name":"http_header_field_manipulation","report_id":"20191028T115649Z_AS28573_eIrzDM4njwMjxBi0ODrerI5N03zM7qQoCvl4xpapTccdW0kCRg","measurement_start_time":"2019-10-28 12:14:07","backend_version":null,"options":[]}
```
bassosimone: can you indent this measurement so the diff review is easier?

af/fastpath/fastpath/tests/data/http_header_field_manipulation_3.json
```diff
@@ -0,0 +1 @@
...ure":"connection_refused_error","request":{"body":null,"headers":{"AccEPt-lAnGuAGe":"en-US,en;q=0.8","accEpT-EncODINg":"gzip,deflate,sdch","ACcepT":"text\/html,application\/xhtml+xml,application\/xml;q=0.9,*\/*;q=0.8","UsEr-AgENT":"Mozilla\/5.0 (Windows; U; Windows NT 5.1; de; rv:1.9.1.7) Gecko\/20091221 Firefox\/3.5.7 (.NET CLR 3.5.30729)","accePt-CHARset":"ISO-8859-1,utf-8;q=0.7,*;q=0.3","hoST":"y1DL3gIKCaEWm0G.com"},"url":"http:\/\/37.218.247.95:80","method":"GeT","tor":{"is_tor":false,"exit_name":null,"exit_ip":null}},"response":null}],"agent":"agent","socksproxy":null},"software_version":"2.3.0","test_runtime":0.950111866,"test_start_time":"2019-10-27 00:20:11","input_hashes":[],"software_name":"ooniprobe","options":[],"data_format_version":"0.2.0","measurement_start_time":"2019-10-27 00:32:01","test_version":"0.2.0","probe_city":null,"test_name":"http_header_field_manipulation","id":"219a713c-7732-4679-97a7-ad71a58ea1ef","input":null,"probe_ip":"127.0.0.1","probe_cc":"PK","probe_asn":"AS45595","annotations":{"platform":"lepidopter"},"test_helpers":{"backend":"http:\/\/37.218.247.95:80"},"report_id":"20191027T002012Z_AS45595_p2qNg0FmL4d2kIuLQXEn36MbraErPPA5i64eE1e6nLfGluHpLk"}
```
bassosimone: can you indent this measurement so the diff review is easier?

bassosimone (approved):
🐳

af/fastpath/fastpath/tests/data/http_header_field_manipulation_2.json
FedericoCeratto: Done

af/fastpath/fastpath/tests/data/http_header_field_manipulation_3.json
FedericoCeratto: Done

FedericoCeratto:
0.57~pr380-122 has been deployed on FSN on 2022-05-28

PR #385, opened 2022-05-17
Fix Telegram spelling

FedericoCeratto:
Create fastpath legacy branch

### DIFF
--- a/af/fastpath/fastpath/core.py
+++ b/af/fastpath/fastpath/core.py
@@ -600,7 +600,7 @@ def score_measurement_telegram(msm):
     scores["http_success_cnt"] = http_success_cnt
     scores["http_failure_cnt"] = http_failure_cnt
     if web_failure is not None:
-        scores["msg"] = "Telegam failure: {}".format(web_failure)
+        scores["msg"] = "Telegram failure: {}".format(web_failure)
     return scores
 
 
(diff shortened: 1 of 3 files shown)


af/fastpath/fastpath/core.py
```diff
@@ -600,7 +600,7 @@ def score_measurement_telegram(msm):
     scores["http_success_cnt"] = http_success_cnt
     scores["http_failure_cnt"] = http_failure_cnt
     if web_failure is not None:
-        scores["msg"] = "Telegam failure: {}".format(web_failure)
+        scores["msg"] = "Telegram failure: {}".format(web_failure)
```
hellais: I suggest we drop the prefix "Telegram failure:" entirely and leave only the web_failure in the message. That way we can make edits to copy inside of explorer directly and add support for translations.

hellais (changes requested):

af/fastpath/fastpath/core.py
FedericoCeratto: Ok, done

hellais: I am suggesting to drop also the "failure: ", otherwise client code still needs to parse that since "failure:" is also a string that needs to be translated to other languages.

In practice what I am suggesting is:
```
scores["msg"] = web_failure
```

FedericoCeratto:
This is a backport to the legacy branch now not necessary anymore. Closing.

PR #386, opened 2022-05-20
Move test helper rotation script to Clickhouse

FedericoCeratto:
Also introduces persistent DNS naming so that the A records do not change during the lifetime of each test helper.
Add support for using multiple zones.
Use CollapsingMergeTree table to support table replication in future.


hellais (approved):
LGTM!

PR #387, opened 2022-06-13
Cleanup, move test list updater to Clickhouse

FedericoCeratto:
Also set dedicated rotation DB role in test-helper rotation tool.


bassosimone (approved):
🐳

PR #393, opened 2022-08-12
Handle the new format of vanilla_tor experiment

### LINKED ISSUE #608: pipeline: vanilla_tor scoring is broken
Here is an example of a measurement of vanilla_tor being blocked: https://explorer.ooni.org/measurement/20220802T215209Z_vanillator_CN_17623_n1_LXFdyo7ao2kE7sQj, but the scores field taken from the API says otherwise: https://api.ooni.io/api/v1/measurement_meta?report_id=20220802T215209Z_vanillator_CN_17623_n1_LXFdyo7ao2kE7sQj.

```
{"anomaly":false,"category_code":null,"confirmed":false,"failure":false,"input":"","measurement_start_time":"2022-08-02T21:55:29Z","measurement_uid":"20220802215530.933188_CN_vanillator_f4e06257ed70aac2","probe_asn":17623,"probe_cc":"CN","report_id":"20220802T215209Z_vanillator_CN_17623_n1_LXFdyo7ao2kE7sQj","scores":"{\"blocking_general\":0.0,\"blocking_global\":0.0,\"blocking_country\":0.0,\"blocking_isp\":0.0,\"blocking_local\":0.0}","test_name":"vanilla_tor","test_start_time":"2022-08-02T21:52:07Z"}
```

hellais:
Fixes: https://github.com/ooni/backend/issues/608


af/fastpath/fastpath/core.py
```diff
@@ -876,17 +876,28 @@ def score_vanilla_tor(msm):
...
+            scores["accuracy"] = 0.0
             scores["msg"] = "Client bug"
             return scores
 
-    tor_log = tk.get("tor_log", None)
-    if tor_log is None:
+    tor_logs = []
+    # Handle incompatibility with legacy test:
+    # https://github.com/ooni/spec/blob/master/nettests/ts-016-vanilla-tor.md#incompatibility-with-ooniprobe-legacy
+    if "tor_logs" in tk and tk["tor_logs"]:
```
FedericoCeratto: You can just use get()

af/fastpath/fastpath/core.py
```diff
@@ -876,17 +876,28 @@ def score_vanilla_tor(msm):
...
+
+    if len(tor_logs) == 0:
         # unknown bug
+        scores["accuracy"] = 0.0
         return scores
 
-    if (
-        "Bootstrapped 100%: Done" in tor_log
-        or "Bootstrapped 100% (done): Done" in tor_log
+    if any(
```
FedericoCeratto: The [] should not be needed, also can you move the strings ouf of the any?

FedericoCeratto (changes requested):

af/fastpath/fastpath/core.py
hellais: How so? I need to distinguish between the case of the test having the key `tor_logs` (which is a list) and the key `tor_log` (which is a string).

af/fastpath/fastpath/core.py
hellais: >The [] should not be needed

ah yes, good call!

> can you move the strings ouf of the any

I'm not sure I understand this suggestion? Should I define the strings as a list and do two nested `any`s or just define the strings as constants and then refer to the constant inside of the current `any` (or something else)?

af/fastpath/fastpath/core.py
FedericoCeratto: (@hellais is on holidays. Implemented in a new commit)

af/fastpath/fastpath/core.py
FedericoCeratto: (@hellais is on holidays. Implemented in a new commit)

FedericoCeratto (approved):

PR #399, opened 2022-09-20
0.66 Implement initial openvpn scoring and obs table

FedericoCeratto:
Also drop PSQL support and dead code.


af/fastpath/fastpath/core.py
```diff
@@ -1429,6 +1424,25 @@ def score_riseupvpn(msm: dict) -> dict:
...
+        tk.get("api_status") == "blocked"
+        or tk.get("ca_cert_status") is False
+        or obfs4 == "blocked"
+        or openvpn == "blocked"
+    )
+    if anomaly:
+        scores["blocking_general"] = 1.0
+
+    scores["extra"] = dict(test_runtime=msm.get("test_runtime"))
+    return scores
```
hellais, line 1431: On what test specification and implementation is this parsing/scoring logic based on?

I checked the spec for the OpenVPN test (https://github.com/ooni/spec/pull/255) and the code for it (https://github.com/ooni/probe-cli/blob/8e26e4993bf48b90a1a89e8a6ada69d6e42b728e/internal/engine/experiment/openvpn/openvpn.go) in the relevant PR: https://github.com/ooni/probe-cli/pull/845, but could not find reference to the keys `transport_status` or `api_status`.

af/fastpath/fastpath/db.py
```diff
@@ -334,3 +228,94 @@ def tf(v: bool) -> str:
...
+        nn(msm, "resolver_asn"),
+        nn(msm, "resolver_ip"),
+        nn(msm, "resolver_network_name"),
+        nn(msm, "software_name"),
+        nn(msm, "software_version"),
+        nn(msm, "success"),
+        tcp_connect_status_success,
+        dget_or(msm, "test_runtime", 0),
+        nn(tk, "transport"),
+    ]
```
hellais: Using this pattern of inserting using a list makes me a little bit nervous. If we add or remove columns we need to be sure that the order of the fields in the row is preserved. It's also very hard to visually validate if the order is correct even currently.

I would suggest using instead the dict based insertion so that you ensure you are putting the values inside of the correct column.

hellais (commented):
I left some comments. I think the most critical point is figuring out what spec/test version this applies to so that a review can be properly carried out.

af/fastpath/fastpath/core.py
FedericoCeratto: The msmt extraction in db.py is based on Ain's feedback and the unit test files provided. The scoring is not implemented yet.

af/fastpath/fastpath/db.py
FedericoCeratto: Updated

ainghazal:
I am updating the spec to match the latest changes in the data format.

af/fastpath/fastpath/tests/test_unit.py
```diff
@@ -518,21 +518,35 @@ def test_score_torsf2():
     }
 
 
+def test_score_openvpn():
+    msm = loadj("torsf_2")
```
hellais: Shouldn't this be loading the `openvpn.json` file here instead of `torsf_2`?

af/fastpath/fastpath/tests/test_unit.py
```diff
@@ -518,21 +518,35 @@ def test_score_torsf2():
...
+
+
+
 # # Bug tests
 
 
 def test_bug_backend351():
     # https://api.ooni.io/api/v1/measurement/temp-id-386770148
-    msm = loadj("bug_351")
+    msm = loadj("openvpn")
```
hellais: Why was this changed to `openvpn` from `bug_351`, but the test is still called `test_bug_backend351`?

af/fastpath/fastpath/core.py
```diff
@@ -1429,6 +1424,25 @@ def score_riseupvpn(msm: dict) -> dict:
     return scores
 
 
+def score_openvpn(msm: dict) -> dict:
+    scores = init_scores()
+    tk = g_or(msm, "test_keys", {})
+    tstatus = tk.get("transport_status") or {}
```
hellais: This scoring logic looks like it's scoring a `riseupvpn` test instead of an `openvpn` one based on the specs I see here:
* https://github.com/ooni/spec/blob/f9bbaa83541484e3e509ffa56dd87b0c5ce8c31a/nettests/ts-026-riseupvpn.md
* https://github.com/ooni/spec/pull/255/commits/2e8887f03fae18b50a8fa94b0ffc0c2f728af3c5

hellais (dismissed):
The scoring logic seems to be for a different test and the provided test fixture is not being used in the tests.

The `openvpn` test result fixture (https://github.com/ooni/pipeline/blob/9b3007f81a786d734df28d984fd3990beefe1f88/af/fastpath/fastpath/tests/data/openvpn.json) upon visual inspection does not seem to be scorable by the logic we see in here.

The unittest is loading the `torsf_2` test json instead of the `openvpn` to test the scoring so the implemented codepath is most likely not being triggered by the test.

af/fastpath/fastpath/core.py
ainghazal: indeed, this seems to be coming from riseupvpn

FedericoCeratto:
As discussed offline, the initial implementation was only about the observation extraction as the test spec was not available yet. score_openvpn and the unit tests are just "stub" code copied from riseupvpn to be replaced. 
Now that the specs are out I'm going to implement the real scoring etc.
Update: the scoring and related tests are implemented.

af/fastpath/fastpath/core.py
FedericoCeratto: Initial basic scoring added.

af/fastpath/fastpath/tests/data/openvpn.json
```diff
@@ -0,0 +1,361 @@
...
+    "Obfuscation=none",
+    "Cipher=AES-256-GCM",
+    "Auth=SHA512",
+    "Compress="
+  ],
+  "probe_asn": "AS4242",
+  "probe_cc": "UK",
+  "probe_ip": "127.0.0.1",
+  "probe_network_name": "UltraLettuce Networks S.L.",
+  "report_id": "",
```
FedericoCeratto, line 24: The report id is currently missing

FedericoCeratto:
0.63~pr399-167 deployed on -test

af/fastpath/fastpath/db.py
```diff
@@ -227,6 +81,45 @@ def _click_create_table_fastpath():
...
 
+def click_create_table_obs_openvpn():
+    sql = """
+    CREATE TABLE IF NOT EXISTS obs_openvpn
+    (
+        anomaly Bool,
+        bootstrap_time Float32,
+        confirmed Bool,
+        error String,
+        failure Bool,
```
hellais: According to the spec this field is a Nullable string

FedericoCeratto:
Bumped version to 0.64 and rebased on https://github.com/ooni/pipeline/pull/401

af/fastpath/fastpath/core.py
FedericoCeratto: It was "stub" code as part of the WIP code, now fixed.

FedericoCeratto:
rebased on master after https://github.com/ooni/pipeline/pull/401 was merged

FedericoCeratto:
Bumped version to 0.64 and rebased on https://github.com/ooni/pipeline/pull/401

ainghazal:
I'm bumping the spec to 0.0.16 after review by @hellais - will send a diff to this PR with any changes.

af/fastpath/fastpath/tests/data/openvpn.json
FedericoCeratto: Test vector updated.

af/fastpath/fastpath/db.py
hellais: https://github.com/ooni/spec/pull/255/files#diff-bf7281a292d49fde7131b3803a9d7f1d7301fa875dec3a907d6594eac299f273R238

af/fastpath/fastpath/db.py
```diff
@@ -227,6 +81,50 @@ def _click_create_table_fastpath():
...
 
 
+def click_create_table_obs_openvpn():
+    sql = """
+    CREATE TABLE IF NOT EXISTS obs_openvpn
+    (
+        anomaly Bool,
+        bootstrap_time Float32,
+        confirmed Bool,
+        error String,
```
hellais: This field does not exist

af/fastpath/fastpath/core.py
```diff
@@ -49,6 +49,7 @@
 import fastpath.portable_queue as queue
 
 import fastpath.utils
+from fastpath.utils import dget_or as g_or
```
hellais, line 52: why? (I mean the `as g_or` part)

af/fastpath/fastpath/db.py
FedericoCeratto: On the DB side failure is a boolean and the value is casted into bool by https://github.com/ooni/pipeline/pull/399/files#diff-ba56dd346dd6041329c79d115e02ad095b974f2bbe6cac021d916fead8e9432fR298

af/fastpath/fastpath/db.py
FedericoCeratto: Removed it, also removed another field that is not used at the moment. We also have a "confirmed" field that currently has no usage for openvpn that we could remove in future.

This PR is meant to be an initial implementation to start gathering measurement and test the usefulness of the obs table. I suggest we improve the code iteratively using CI/CD as longs as we are not tying our hands by publishing API to end users (that would need to remain stable or backward-compatible).

af/fastpath/fastpath/core.py
FedericoCeratto: It's a simple getter. The code is moving toward lenient getters that can handle null values or missing keys - and adding unit tests. At the end of the migration we can do some refactoring.

af/fastpath/fastpath/db.py
hellais: I think that's quite confusing and much less useful than storing the actual failure string. You can always get the bool from checking if the string is not NULL.

FedericoCeratto: Updated to use a string.

af/fastpath/fastpath/core.py
hellais: Right, but why are you aliasing it to `g_or` instead of using the `dget_or` name as is done in other parts of the code?

FedericoCeratto: It was written like that earlier on elsewhere and I wanted to avoid mixing too much feature development and code cleanup in the same PR.

FedericoCeratto:
Rebased after merge of the Whatsapp PR.

hellais (approved):
LGTM

FedericoCeratto:
0.66~pr399-188 deployed on test

FedericoCeratto:
0.66~pr399-188 deployed on prod

PR #402, opened 2022-11-21
Extract blocking type

FedericoCeratto:
Analysis on:
https://jupyter.ooni.org/notebooks/notebooks/2022%20%5Bfederico%5D%20blocking_type.ipynb


af/fastpath/fastpath/db.py
```diff
@@ -182,24 +184,25 @@ def tf(v: bool) -> str:
...
+        probe_asn=asn,
+        test_name=test_name,
+        test_start_time=test_start_time,
+        measurement_start_time=measurement_start_time,
+        scores=ujson.dumps(scores),
+        platform=nn(msm, "platform"),
+        anomaly=tf(anomaly),
+        confirmed=tf(confirmed),
+        msm_failure=tf(msm_failure),
+        blocking_type=nn(msm, "blocking"),
```
hellais: Shouldn't this be extracted from the `analysis` field of the `scores`? Measurements don't have as a top level key the value `blocking`.

af/fastpath/fastpath/db.py
```diff
@@ -182,24 +184,25 @@ def tf(v: bool) -> str:
...
+        measurement_uid=measurement_uid,
+        report_id=nn(msm, "report_id"),
+        input=input_,
+        probe_cc=nn(msm, "probe_cc"),
+        probe_asn=asn,
+        test_name=test_name,
+        test_start_time=test_start_time,
+        measurement_start_time=measurement_start_time,
+        scores=ujson.dumps(scores),
+        platform=nn(msm, "platform"),
```
hellais: This should be reading the value of the keyword argument `platform` being passed to the function instead of the key in the measurement.

hellais: This looks like a regression bug that was introduced during the migration to clickhouse, so it's not something new. In postgres we were extracting it properly: https://github.com/ooni/pipeline/blob/master/af/fastpath/fastpath/db.py#L131 and I confirmed we have this always set to the empty string in the DB.

af/fastpath/fastpath/db.py
FedericoCeratto: Updated after some commits. It's from test_keys -> blocking and it's directly from the measurement so that it will support any compatible test in future

af/fastpath/fastpath/db.py
FedericoCeratto: Fixed in the last commit.

af/fastpath/fastpath/db.py
hellais: I don't think we should be moving in the direction of putting more analysis logic inside of the probe and relying on them to generate a valid blocking key.

It should be the responsibility of the analysis engine to come up with a suitable blocking value based on the results of the measurement (only the backend is aware of the blocking fingerprints and we can improve the blocking heuristics iteratively).

hellais (changes requested):
Based on the comment https://github.com/ooni/pipeline/pull/402#discussion_r1028344432 it seems like this is going in the direction of requiring us to modify tests to generate a `blocking` key inside of the probe, which I don't think is the direction where we should be going for.

Rather I would like us to have the value of this key be determined by analysis logic that lives inside of the backend which we can improve upon without requiring changes to the experiments. This would mean as we improve the analysis logic we will be able to improve the accuracy of results.

Moreover doing it in the probe, means that we are analysis outcomes from potentially different software versions that can lead to inconsistent comparisons making it hard to identify if the results are consistent over time.

hellais (approved):
LGTM

We proceed with merging this and then backfill of the column.

Ensure the counts are consistent during the backfill or if not possible display a warning on explorer saying the counts are wrong.

FedericoCeratto:
0.69~pr402-227 deployed on -test

FedericoCeratto:
0.69~pr402-227 deployed on FSN

PR #405, opened 2022-12-19
Move backup to Clickhouse

FedericoCeratto:
Temporarily use CSV as the native backup function is not available. (hardcode table names for now).
Chunk tables.
Move backup code to dedicated module.
The CSV export function can be used to provide tables to 3rd party researchers.
Leave unused code based on native backup for future use.
Implement the initial version described in https://docs.google.com/document/d/1_uFMSlh0wkd9VNBmjrbEUJRsY6-sVD7kHVJAuzp7Lng


af/analysis/ooni_db_backup.py
```diff
@@ -0,0 +1,287 @@
...
+    assert s_blob.startswith("measurement_uid String, ")
+    # s_colnames is a string to be placed into ##colnames## without quotes
+    basesql = """
+    INSERT INTO FUNCTION s3(%(s3path)s, %(aws_id)s, %(aws_secret)s,
+        'CSVWithNames',
+        %(s_blob)s
+    )
+    SELECT ##colnames##
+    FROM fastpath
+    WHERE toYYYYMM(measurement_start_time) = %(yyyymm)s
```
hellais: I think the tricky thing here is that people can upload old measurements later, so if at a certain time you have backed up data from a given month, you will have to back it up again every time you run the backup process because you don't have the guarantee that other measurements have been uploaded in the meantime.

How about instead you use the report_id or measurement_uid prefixes for determining the partitions of the csv backup files?

If you decide to keep it like this, you may still want to do the partitioning into files at the clickhouse level instead of wrapping it in a for loop, like this: https://clickhouse.com/docs/en/integrations/s3/s3-table-functions/#splitting-large-files.

af/analysis/ooni_db_backup.py
```diff
@@ -0,0 +1,287 @@
...
+    tblname = "fastpath"
+    s_colnames, s_blob = describe_table(click, tblname)
+    assert s_blob.startswith("measurement_uid String, ")
+    # s_colnames is a string to be placed into ##colnames## without quotes
+    basesql = """
+    INSERT INTO FUNCTION s3(%(s3path)s, %(aws_id)s, %(aws_secret)s,
+        'CSVWithNames',
+        %(s_blob)s
+    )
+    SELECT ##colnames##
```
hellais: What is the reason for expliciting the column names instead of just passing in `*`?

~Using `*` seems more robust to potential schema changes.~

I see now that the column names are being derived from the DESCRIBE query, yet it's still unclear to me what the benefit is of doing that vs selecting `*`

af/analysis/ooni_db_backup.py
```diff
@@ -0,0 +1,287 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+OONI DB backup tool
+Configured with /etc/ooni/db-backup.conf - Runs as a system daemon
```
hellais, line 5: Is it possible to have an example `db-backup.conf` file checked in?

af/analysis/ooni_db_backup.py
```diff
@@ -0,0 +1,287 @@
...
+        s3path = f"{baseurl}/csv/{tblname}_{chunk_num:05}.csv.zstd"
+        kw = dict(
+            aws_id=aws_id,
+            aws_secret=aws_secret,
+            s3path=s3path,
+            limit=chunk_size,
+            offset=chunk_num * chunk_size,
+            s_blob=s_blob,
+        )
+        sql = "SELECT count() FROM jsonl LIMIT %(limit)s OFFSET %(offset)s"
```
hellais: If I understand the intent of this function correctly, you are trying to export the jsonl table in several chunks of 100M rows and use this `count()` query to understand if you are at the last chunk.

I believe there are however two bugs atm:

1. the `basesql` query does not make use of `LIMIT` and `OFFSET` so it's effectively just selecting and copying the whole table on the first iteration of the `while True` loop
2. the above `SELECT count()` query will always return empty rows on the second interation, because `LIMIT` and `OFFSET` constraints are being applied on the result of the `count()` query.

If you want to know the count of the current chunk, you need an extra level of nesting in the SELECT query:
```
SELECT COUNT() FROM (
    SELECT measurement_uid FROM jsonl LIMIT %(limit)s OFFSET %(offset)s
)
```

That said, I'm not sure this is the best approach, as on every iteration you are basically scanning all the rows until the offset at every iteration of the loop.

It's probably better to make use of the `PARTITION BY` clause, which I would suggest doing on the `date` (AFAIK `date` is what we used to call `bucket_date`) column which we know is going to always increase and will not contain new data going forward.

af/analysis/ooni_db_backup.py
FedericoCeratto: > if at a certain time you have backed up data from a given month,

The initial version of the tool is not designed to do incremental backups. When we switch to a newer version of Clickhouse we can use its native, incremental backup support.

> How about instead you use the report_id or measurement_uid prefixes for determining the partitions of the csv backup files?

This is a backup tool meant to run against our only production database. The table is indexed by measurement_start_time and we have to follow the index if we want to minimize the CPU and I/O load on the database.
Additionally, partitioning the files by YYYYMM makes it easier to understand the contents and reuse them.
Also, native partitioning by clickhouse does not seem to be supported by the current version.

af/analysis/ooni_db_backup.py
FedericoCeratto: It is required by the current version of Clickhouse

af/analysis/ooni_db_backup.py
hellais: Ah that's unfortunate that partioned writes are not supported in the version of clickhouse we are using.

According to the changelog support for it was introduced in v21.10 (https://clickhouse.com/docs/en/whats-new/changelog/2021/#clickhouse-release-v2110-2021-10-16).

> Additionally, partitioning the files by YYYYMM makes it easier to understand the contents and reuse them.

My suggestion was about using the prefix from the report_id or measurement_id, so you would effectively end up with something which is the same as using the `measurement_start_time`, except that the content of a particular file will not change over time.

I do understand that there are load constraints to take into account and I guess it's OK for the time being to use this approach.

af/analysis/ooni_db_backup.py
FedericoCeratto: Added.

af/analysis/ooni_db_backup.py
FedericoCeratto: That's correct. The function was still being finished and tested. Now it's rewritten to use a temporary file due to a bug in Clickhouse in the uploading to S3.

FedericoCeratto:
Analysis package version bumped to 1.4.5

hellais (approved):
LGTM! 🐳

PR #407, opened 2023-02-09
Use fingerprints from github

FedericoCeratto:
Implements the basic version as in https://docs.google.com/document/d/1QvZGIBOHcratnM7DwJCHtlmshwuMlXuF9ZGPavjWVyA

Switch fingerprint matching to a new logic. Add matching fingerprints to the scores struct.


FedericoCeratto:
analysis package version bumped to 1.5

af/fastpath/fastpath/core.py
```diff
@@ -59,7 +58,28 @@
...
+
+class Fingerprint(TypedDict):
+    name: str
+    scope: str
+    other_names: str
+    location_found: str
+    pattern_type: str
+    pattern: str
+    confidence_no_fp: int
+    expected_countries: str
```
hellais: I think the metadata we only really care to have in the scores dict is:
* name
* scope
* location_found
* confidence_no_fp
* expected_countries

The others are probably OK to exclude

af/fastpath/fastpath/core.py
```diff
@@ -1248,71 +1276,27 @@ def score_http_requests(msm: dict) -> dict:
...
+    matches: List[Dict] = []
+    for req in requests:
+        is_tor = gn(req, "request", "tor", "is_tor")
+        resp = g_or(req, "response", {})
+        body = gn(req, "response", "body")
+        match_http_body_fingerprints(resp, matches)
+        match_http_headers_fingerprints(resp, matches)
+
+    if matches:
+        if is_tor:
```
hellais: I'm not sure this logic is correct.

The `is_tor` key can be set on any individual requests. If it's set you should not flag the whole measurement as inaccurate.

Rather, you should not perform any matching on the fingerprint when it's is_tor, since that's not going to be related to the metadata of the probe itself.

af/fastpath/fastpath/core.py
```diff
@@ -1248,71 +1276,27 @@ def score_http_requests(msm: dict) -> dict:
...
-            prefix = fp["header_prefix"]
-            if name in headers and headers[name].startswith(prefix):
-                if is_tor:
-                    scores["accuracy"] = 0.0
-                    log.debug(f"Failed measurement {rid} {inp}")
-                    return scores
-                scores["confirmed"] = True
-                log.debug("matched header prefix %s %r", msm_cc, prefix)
+        else:
+            scores["confirmed"] = True
```
hellais, line 1330: This should be set only when `probe_cc` of the measurement is in the `expected_countries` of any matched fingerprint OR scope in any of the matching fingerprints is one of `[isp, nat]`

af/fastpath/fastpath/core.py
```diff
@@ -904,22 +940,31 @@ def score_vanilla_tor(msm: dict) -> dict:
...
-        scores[l] += 1.0
+    if matches:
+        scores["fingerprints"] = [minifp(fp) for fp in matches]
+
+    probe_cc = g_or(msm, "probe_cc", "ZZ").upper()
+    for fp in matches:
+        if (
+            probe_cc in fp["expected_countries"]
+            or not fp["expected_countries"]
+            or fp["scope"] in ("isp", "nat")
```
hellais: This logic should be:
```
        if (
            (probe_cc in fp["expected_countries"] and len(fp["expected_countries"]) > 0)
            or fp["scope"] in ("isp", "nat")
        )
```

af/fastpath/fastpath/core.py
hellais: This code seems to be duplicate of what is here: https://github.com/ooni/pipeline/pull/407/files#r1118514733. It's probably best to refactor this into a dedicated function so if we make changes to the logic (as we improve our understanding of how accurate the fingerprints are), we can update it only in one place and not forget the other.

af/fastpath/fastpath/core.py
FedericoCeratto: Added.

af/fastpath/fastpath/core.py
FedericoCeratto: Done

FedericoCeratto:
fastpath `0.70~pr407-228` and analysis `1.5~pr407-229` deployed on test

FedericoCeratto:
New dashboard: https://grafana.ooni.org/d/JNlK8ox4z/fingerprints?orgId=1&from=now-12h&to=now&refresh=5m

af/fastpath/fastpath/core.py
FedericoCeratto: Discussed on Slack and updated.

af/fastpath/fastpath/core.py
FedericoCeratto: Part of the code is duplicated on purpose in order to merge it later on based on how it evolves in the meantime. Part of it has been deduplicated into match_http_body_fingerprints and match_http_headers_fingerprints

af/fastpath/fastpath/core.py
```diff
@@ -285,96 +303,114 @@ def process_measurements_from_s3() -> None:
...
+    if body is None:
+        return
+
+    for fp in fingerprints["http"]:
+        if fp["location_found"] != "body":
+            continue
+        tb = time.time()
+        bm = fp["pattern"]
+        if isinstance(body, bytes):
+            idx = body.find(bm.encode())
```
hellais, line 338: This will lead us to missing some fingerprints in the event of response bodies in some non-utf8 encoding. I have seen this happen before and is the reason why I have to do this quite sketching thing in ooni/data: https://github.com/ooni/data/blob/main/oonidata/datautils.py#L74.

Doing this though means we have to pay a pretty high tax for decoding all bodies, so probably we should not do this in the fastpath, but I thought it would still be useful to mention in this context as something to keep in mind.

af/fastpath/fastpath/core.py
```diff
@@ -1679,35 +1696,55 @@ def core():
         clean_caches()
 
 
-def setup_fingerprints():
-    """Setup fingerprints lookup dictionary
-    "ZZ" applies a fingerprint globally
-    """
-    # pre-process fingerprints to speed up lookup
+def extract_expected_countries(ec):
```
hellais, line 1699: Is this function still being used? I see it only mentioned in the unittests.

af/fastpath/fastpath/core.py
```diff
@@ -1679,35 +1696,55 @@ def core():
...
+    return sorted(ecs)
+
+
+@metrics.timer("prepare_fingerprints")
+def prepare_fingerprints(dns_fp, http_fp):
+    for fp in dns_fp + http_fp:
+        exp = set(fp["expected_countries"].strip().split(","))
+        exp.discard("")
+        # Discard ZZ as we don't handle it currently
+        exp.discard("ZZ")
```
hellais: I think that since in 
```
def is_confirmed(probe_cc: str, fp: Fingerprint) -> bool:
    """Checks if a matching fingerprint represents confiremed blocking"""
    if probe_cc != "ZZ" and probe_cc in fp["expected_countries"]:
```

You can probably drop these two lines where you discard `ZZ`. Once the PR cleaning up the list is, probably the same thing can be done for the previous discard.

hellais (approved):
Left a few very minor comments, but other LGTM! 🐳

af/fastpath/fastpath/core.py
FedericoCeratto: I'm doing the discard/ignore twice on purpose because it has different meaning (cleaning up the global variable VS ignoring msmts in ZZ), given that we are planning to change the matching criteria in future.

af/fastpath/fastpath/core.py
FedericoCeratto: I'd rather keep it around to make the code testable in case we add more clever logic in future.
Refactored to use it. (FWIW)

FedericoCeratto:
analysis 1.5~pr407-233 deployed on FSN

FedericoCeratto:
fastpath 0.70~pr407-233 deployed on FSN

PR #412, opened 2023-03-02
Rotation keep history

FedericoCeratto:
Add basic testing, cleanups

### DIFF
--- a/af/analysis/rotation.py
+++ b/af/analysis/rotation.py
@@ -1,9 +1,15 @@
 #!/usr/bin/env python3
 """
-Spaws and rotates hosts on Digital Ocean. Runs a setup script on the host
-at first boot.
+Spaws and rotates hosts on Digital Ocean.
+- Select datacenters and spawn VMs
+- Runs a setup script on the host at first boot
+- Keeps a list of live and old hosts in a dedicated db table
+- Create SSL certificates using the Digital Ocean API
+- Performs end-to-end test on newly created VMs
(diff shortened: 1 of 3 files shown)


af/analysis/tests/test_rotation.py
```diff
@@ -0,0 +1,159 @@
...
+import datetime
+from unittest.mock import Mock, create_autospec, call
+
+import pytest  # debdeps: python3-pytest
+
+import rotation as ro
+
+
+@pytest.fixture(autouse=True, scope="session")
+def mock_everything():
```
bassosimone, line 14: I love the name of this function! 🤣

bassosimone (approved):
🐳

af/analysis/rotation.py
```diff
@@ -24,6 +30,7 @@
     `ipaddr` IPv4,
     `ipv6addr` IPv6,
     `draining_at` Nullable(DateTime('UTC')),
+    `destroyed_at` Nullable(DateTime('UTC')),
     `sign` Int8
 )
 ENGINE = CollapsingMergeTree(sign)
```
hellais, line 36: If we are effectively always using sign = 1, why don't we at that point just use a ReplacingMergeTree?

FedericoCeratto: In the past we wanted to control deletions using CollapsingMergeTree. In future versions we might switch database entirely as discussed.

PR #415, opened 2023-03-15
Increase worker count

FedericoCeratto:


af/fastpath/fastpath/localhttpfeeder.py
```diff
@@ -35,9 +34,8 @@ def handler_app(environ, start_response):
             assert path.startswith("/2")
             msmt_uid = path[1:]
             data = environ["wsgi.input"].read()
-            while queue.qsize() >= 5000:
-                time.sleep(0.1)
-            queue.put((data, None, msmt_uid))
+            msm_tup = (data, None, msmt_uid)
+            queue.put(msm_tup, block=True)
```
hellais, line 38: Should we perhaps put a timeout in the blocking put call and log when we reach the timeout, to signal that we have reached the maximum size of the queue?

Something like:
```
while True:
    try:
         queue.put(msm_tup, block=True, timeout=0.5)
         break
    except queue.Full:
         # XXX log somewhere that the queue is full
```

hellais:
Great to see that the portable_queue was dropped in favour of limiting the queue size in the constructor!

I left a comment about potentially adding some logging to the blocking put call.

FedericoCeratto:
Moved to the new repo.

